PR #102, opened 2015-01-25
Added support for Scanning []interface in a struct.

norcalli:
When using ScanSlice, this change allows having a `[]interface` in a struct. Otherwise, it gives a message that seems like it should be wrong:

```
redigo: Scan cannot convert from []interface {} to []interface {}
```

I also added recursive slicing otherwise.


garyburd:
Please provide an example.

norcalli:
Example of when I ran into this? I could dig up one in a bit.

garyburd:
No response, closing PR.

PR #112, opened 2015-03-01
Scan HGETALL response to map

TheAJ:
A helper function to map HGETALL response to a map[string]sting type.

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -287,3 +287,26 @@ func Ints(reply interface{}, err error) ([]int, error) {
 	}
 	return ints, nil
 }
+
+// StringMap is a helper that converts an array of strings (alternating key, value)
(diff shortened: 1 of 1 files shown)


garyburd:
The function below is more convenient to use. It should be added to reply.go.

    func StringMap(result interface{}, err error) (map[string]string, error) {
       values, err := Values(result, err)
       if err != nil {
         return nil, err
      }  
      if len(src)%2 != 0 {
        return errors.New("redigo: StringMap expects even number of values result")
      }
      m := make(map[string]string, len(src)/2)
      for i := 0; i < len(src); i += 2 {
          key, okKey := src[i].([]byte)
         value, okValue := src[i+1].([]byte)
          if !okKey || !okValue {
               return nil, errors.New("redigo: ScanMap key not a bulk string value")
          }
          m[string(key)] = string(value)
      }
      return nil
    }


}

PR #126, opened 2015-06-23
add IntMap Int64Map

terryh:


garyburd:
Squash the commits.  Remove the reference to CONFIG GET from the documentation.  This command returns non-integer values.

terryh:
Yes, squash to one commit.

redis/reply.go
```diff
@@ -310,3 +310,49 @@ func StringMap(result interface{}, err error) (map[string]string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	if len(values)%2 != 0 {
+		return nil, errors.New("redigo: IntMap expects even number of values result")
+	}
+	m := make(map[string]int, len(values)/2)
+	for i := 0; i < len(values); i += 2 {
+		key, okKey := values[i].([]byte)
+		value, okValue := Int(values[i+1], nil)
```
garyburd: It is not conventional in Go to name errors with the prefix 'ok'.  Rename 'okValue' to 'err'.  Also, return the error because it's informative. 

        key, ok := values[i].([]byte)
        if !ok {
            return nil, errors.New("redigo: ScanMap key not a bulk string value")
        }
        value, err := Int(values[i+1], nil)
        if err != nil {
           return nil, err
       }

terryh:
You are right, change to your update.

PR #127, opened 2015-06-24
Improve performance of LookupCommandInfo.

clutchski:
Converting command names to upper case is a non-trival part of our
profile. Assuming that 99% of programmers don't use mixed case,
we can optimize that by explicitly checking the lower/upper case
commands and short-circuiting before upper casing.

This makes mixed cases marginally slower, but makes the common case much
faster. A good trade-off I think.

```
BEFORE:

BenchmarkLookupCommandInfoCorrectCase    2000000           716 ns/op
BenchmarkLookupCommandInfoMixedCase      3000000           532 ns/op

AFTER:

BenchmarkLookupCommandInfoCorrectCase   10000000           131 ns/op
BenchmarkLookupCommandInfoMixedCase     2000000           593 ns/op
```

### DIFF
--- a/internal/commandinfo.go
+++ b/internal/commandinfo.go
@@ -29,17 +29,49 @@ type CommandInfo struct {
 	Set, Clear int
 }
 
+var (
+	watch      = CommandInfo{Set: WatchState}
+	unwatch    = CommandInfo{Clear: WatchState}
+	multi      = CommandInfo{Set: MultiState}
+	exec       = CommandInfo{Clear: WatchState | MultiState}
+	discard    = CommandInfo{Clear: WatchState | MultiState}
+	psubscribe = CommandInfo{Set: SubscribeState}
+	subscribe  = CommandInfo{Set: SubscribeState}
+	monitor    = CommandInfo{Set: MonitorState}
+)
+
 var commandInfos = map[string]CommandInfo{
-	"WATCH":      {Set: WatchState},
-	"UNWATCH":    {Clear: WatchState},
-	"MULTI":      {Set: MultiState},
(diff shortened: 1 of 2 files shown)


garyburd:
Take a look at https://github.com/garyburd/redigo/compare/lookup-perf.  I think it will match the performance of this PR, but be more maintainable when all commands are added to support Redis cluster.

clutchski:
:+1: as long as the benchmark looks good, seems like a fine solution to me.

clutchski:
also .. i have some crap in the benchmark file (e.g. unrelated comments) and a wrong err msg which i just realized should be cleaned up.

garyburd:
Perf is identical for matching case.

    $ git checkout clutchski-performance
    Switched to branch 'clutchski-performance'
    $ go test -test.bench=.
    PASS
    BenchmarkLookupCommandInfoCorrectCase   20000000           104 ns/op
    BenchmarkLookupCommandInfoMixedCase  3000000           519 ns/op
    ok      github.com/garyburd/redigo/internal 4.276s
    $ git checkout lookup-perf
    Switched to branch 'lookup-perf'
    $ go test -test.bench=.
    PASS
    BenchmarkLookupCommandInfoCorrectCase   20000000           102 ns/op
    BenchmarkLookupCommandInfoMixedCase  3000000           535 ns/op
    ok      github.com/garyburd/redigo/internal 4.293s

I'll cleanup the comments and submit it.

clutchski:
thank you sir.

garyburd:
Thank you for pointing out the issue.  See 27a6a7d7c76db7d75feb28f82e91cb34160606c3.

PR #132, opened 2015-07-13
In this case, the reply is []strings{string,string...}, the func Strings does not handle this case. Fix it.

cfxks1989:
In this case, the reply is []strings{string,string...}, the func Strings does not handle this case.
```
// test_redigo project main.go
package main

import (
	"fmt"
	redis "github.com/garyburd/redigo/redis"
)

const (
	MULTI = "MULTI"
	TYPE  = "TYPE"
	HSET  = "HSET"
	SADD  = "SADD"
	SET   = "SET"
	LPUSH = "LPUSH"
	EXEC  = "EXEC"
)

func main() {

	r, _ := redis.Dial("tcp", "127.0.0.1:6379")

	r.Send(MULTI)
	r.Send(SADD, "SETKEY", "one")
	r.Send(LPUSH, "LISTKEY", "list")
	r.Send(HSET, "HASHKEY", "key", "value")
	r.Send(SET, "STRINGKEY", "string")
	r.Do(EXEC)

	r.Send(MULTI)
	r.Send(TYPE, "SETKEY")
	r.Send(TYPE, "LISTKEY")
	r.Send(TYPE, "HASHKEY")
	r.Send(TYPE, "STRINGKEY")

	fmt.Println(redis.Strings(r.Do(EXEC)))
```

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -256,11 +256,14 @@ func Strings(reply interface{}, err error) ([]string, error) {
 			if reply[i] == nil {
 				continue
 			}
-			p, ok := reply[i].([]byte)
-			if !ok {
+			switch reply[i].(type) {
+			case []byte:
+				result[i] = string(p)
+			case string:
+				result[i] = string(p)
+			default:
 				return nil, fmt.Errorf("redigo: unexpected element type for Strings, got type %T", reply[i])
 			}
-			result[i] = string(p)
 		}
 		return result, nil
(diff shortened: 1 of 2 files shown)


garyburd:
Use [redis.Values](http://godoc.org/github.com/garyburd/redigo/redis#Values) in the cases where the response is not guaranteed to be an array of bulk strings.

PR #136, opened 2015-08-06
merge 

blackbeans:


garyburd:
I am rejecting this PR because it will break applications that call the NewPool function.

Applications should initialize the pool fields directly to set MaxActive and IdleTimeout. There's no need to use the NewPool helper function.

PR #137, opened 2015-08-11
Export Dialer interface for Redis URLs

shosti:
Let me know if this interface is what you had in mind. Ideally, the
NewDialer function should probably also accept some other URL schemes as
well (let me know your thoughts--I think the Ruby client also accepts
unix and TCP sockets).

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -21,7 +21,10 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/url"
+	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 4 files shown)


garyburd:
Here's the API that I was expecting:

    type Dialer struct {
	// NetDial specifies the dial function for creating TCP connections. If
	// NetDial is nil, then net.Dial is used.
	NetDial func(network, addr string) (net.Conn, error)

	// ReadTimeout specifies the timeout for reading a single command
	// reply. If ReadTimeout is zero, then no timeout is used.
	ReadTimeout time.Duration

	// WriteTimeout specifies the timeout for writing a single command.  If
	// WriteTimeout is zero, then no timeout is used.
	WriteTimeout time.Duration
    }

    func (d *Dialer) DialURL(u string) (Conn, error) {
    ...
    }

    var DefaultDialer = &Dialer{ ... }

A key point of this design is to specify options using a struct instead of function arguments.  This makes it possible to add new options without adding more variations of the Dial function.

garyburd:
After looking at the new Gorilla CSRF package this morning, I think that option functions are a better way to handle options.  Rob Pike [wrote about option functions on his blog](http://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html).

The API should look something like this:

     type DialOption struct { f func(*someInternalType) }
     func ReadTimeout(d time.Duration) DialOption
     func WriteTimeout(d time.Duration) DialOption
     func DialURL(u string, options ...DialOption) (Conn, err)

I will create a new DialOptions function. From there, it should be easy to create DialURL.

garyburd:
I added dial options on a branch. See https://github.com/garyburd/redigo/commit/02b5249fe0eaf19f4c875e7b7b8b542202393396.

I am uncertain about how to name the options functions.  The current names are clunky, but removing the DialOption prefix may cause clashes with option functions for other features. I'll resolve the name issue soon, but the work so far should unblock work on the Redis URL feature.

shosti:
Awesome, thanks for keeping me posted! One question: couldn't you just replace `DialWithOptions(...)` with just `Dial(network, address string, options ...DialOption)` (and possibly deprecate `DialTimeout`)? It should be backward-compatible, and would make the API simpler I think. (Also, my vote from the peanut gallery would probably be to name the option functions `DialReadTimeout`, `DialWriteTimeout` and so forth, but whatever you decide.)

garyburd:
DialWithOptions and DIal are not backward compatible for applications that use a function pointer to Dial.  I agree that adding yet another Dialxxx function complicates that API. I'll think about this more.

shosti:
Oh, OK, didn't think of that, good to know.

garyburd:
- Use updated `Dial` function with `DialOption`s.
- Use [net.SplitHostPort](http://godoc.org/net#SplitHostPort) to split the host and port. If SplitHostPort returns an error, assume that hostPort contains host name only.

shosti:
I think it will be easier to make a new PR with the new interface. Closing.

PR #139, opened 2015-08-13
Added GetWait(timeout) method to the Pool to have a chance of timing …

### LINKED ISSUE #105: Blocking with timeout when Get PooledConn
## Please support new features : 
- We need blocking with timeout when Pool.Get() return PooledConn.
#### Manual setting

   The timeout can be manual configured.

tegioz:
…out when getting a connection from the pool

- Added method GetWait(timeout time.Duration) using one of the options suggested in #105 
- Pool is now using a channel instead of a condition variable as internal synchronisation mechanism for goroutines waiting for a pool connection (https://github.com/golang/go/issues/9578)
- Added a mutex to poolTestConn and poolDialer helpers in pool tests as "go test -race" wasn't happy  (races were just in tests code, nothing important)
- Added TestWaitPoolTimeout to test the new GetWait method, startGoroutines function in pool tests now accepts a timeout as well passed to GetWait
- All tests passing, benchmarks numbers almost identical

This should close #105.

Hope it fits ok. Please let me know if you need anything else or if you find any problem.

Cheers.-

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -35,8 +35,15 @@ var nowFunc = time.Now // for testing
 // pool has been reached.
 var ErrPoolExhausted = errors.New("redigo: connection pool exhausted")
 
+// ErrTimeout is returned from a pool connection method when the wait to get a
+// connection from the pool times out.
+var ErrTimeout = errors.New("redigo: get connection from pool timed out")
+
+// ErrPoolClosed is returned from a pool connection method when the pool was
+// closed while waiting to get a connection from the pool.
+var ErrPoolClosed = errors.New("redigo: pool closed while waiting on get")
+
 var (
-	errPoolClosed = errors.New("redigo: connection pool closed")
 	errConnClosed = errors.New("redigo: connection closed")
 )
 
@@ -120,9 +127,11 @@ type Pool struct {
 	// for a connection to be returned to the pool before returning.
 	Wait bool
 
+	// Channel used to synchronize goroutines waiting to get a pool connection.
+	waitCh chan int
+
 	// mu protects fields defined below.
 	mu     sync.Mutex
-	cond   *sync.Cond
 	closed bool
 	active int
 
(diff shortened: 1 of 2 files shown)


redis/pool.go
```diff
@@ -120,9 +127,11 @@ type Pool struct {
 	// for a connection to be returned to the pool before returning.
 	Wait bool
 
+	// Channel used to synchronize goroutines waiting to get a pool connection.
+	waitCh chan int
```
garyburd: This field is protected by mu.  Move below declaration of mu so that comment on mu applies.

garyburd:
The wait test failed: https://travis-ci.org/garyburd/redigo/jobs/75388867#L176

Change https://github.com/tegioz/redigo/blob/83ae7711b67cf54e16749db11b11e0b030474a4f/redis/pool.go#L256 to return ErrPoolClosed.

tegioz:
Hi,

Will change those details and send you a new pull request, thanks for the prompt review :)

I've seen the Travis error in just one of the builds, but I haven't managed to see it locally, I'll have another look though.

garyburd:
I think that [swapping the order of these lines](https://github.com/tegioz/redigo/blob/83ae7711b67cf54e16749db11b11e0b030474a4f/redis/pool_test.go#L450-L451) will fix the flaky test.

garyburd:
There is a race. 

Goroutine G is getting a connection from the pool. Goroutine P is returning a connection to the pool.

[G](https://github.com/tegioz/redigo/blob/83ae7711b67cf54e16749db11b11e0b030474a4f/redis/pool.go#L284-L285): p.active is equal to p.maxActive.  G releases lock before waiting.

		p.mu.Unlock()
		if timeout > 0 {

[P](https://github.com/tegioz/redigo/blob/83ae7711b67cf54e16749db11b11e0b030474a4f/redis/pool.go#L202): P decrements active count ...

	p.active -= 1
	p.signalWaiting()

[P](https://github.com/tegioz/redigo/blob/83ae7711b67cf54e16749db11b11e0b030474a4f/redis/pool.go#L327-L332):  ... and signals a waiting goroutine.  G is not waiting yet.

	if !p.closed {
		select {
		case p.waitCh <- 1:
		default:
		}
	}

[G](https://github.com/tegioz/redigo/blob/83ae7711b67cf54e16749db11b11e0b030474a4f/redis/pool.go#L286-L293): G begins wait for p.active < p.MaxActive.  P has already signaled, so G waits until another connection is returned to the pool.

			select {
			case _, ok := <-p.waitCh:
				if !ok {
					return nil, ErrPoolClosed
				}
			case <-time.After(timeout):
				return nil, ErrTimeout
			}

One possible fix is to use a channel as a counting semaphore instead of using a counter a channel as a mutex.

tegioz:
Closing this pull request, will send a new one shortly including the suggested changes.

PR #141, opened 2015-08-14
Added GetWait(timeout) method to Pool

tegioz:


garyburd:
Consider the scenario when Get is called when p.active == p.MaxActive.  

A new `semaphore` is allocated with an empty `counter` channel. 

		if p.sem == nil {
			if p.MaxActive > 0 {
				p.sem = newSemaphore(p.MaxActive)
			} else {
				p.sem = newSemaphore(defaultMaxActive)
			}
		}

Because the `counter` channel is empty, the call to wait returns immediately with `err` == nil.

		p.mu.Unlock()
		if err := p.sem.wait(timeout); err != nil {
			return nil, err
		}
		p.mu.Lock()

The for loop executes `MaxActive` times before blocking. 

I stopped looking at the code here. Even if the pool does work correctly, this is too subtle. The value of p.active and len(p.sem.conter) must stay in sync for correct operation, but it's not obvious that they do.

tegioz:
Hope the last commit helps. Happy to submit it again in a single commit after fixing the conflicts.

garyburd:
I am very busy for the next couple of weeks. I'll continue reviewing this in September.

tegioz:
Ok, thanks :)

garyburd:
The application can wait for a multiple of timeout (2 * timeout, 3 * timeout, ....).

garyburd:
The function should be rewritten to replace the `active` counter with a `chan struct{}`.  Retrofitting the condition variable code to use channels is complex.  

I'll do this in a week or two.

PR #142, opened 2015-08-15
Be more descriptive about conversion errors

connor4312:
This wraps conversion errors to add a more descriptive and easier to track down message.

### DIFF
--- a/redis/scan.go
+++ b/redis/scan.go
@@ -367,7 +367,7 @@ func ScanStruct(src []interface{}, dest interface{}) error {
 			continue
 		}
 		if err := convertAssignValue(d.FieldByIndex(fs.index), s); err != nil {
-			return err
+			return errors.New("redigo: error converting '" + string(name) + "' in ScanStruct: " + err.Error())
 		}
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


redis/scan.go
```diff
@@ -367,7 +367,7 @@ func ScanStruct(src []interface{}, dest interface{}) error {
 			continue
 		}
 		if err := convertAssignValue(d.FieldByIndex(fs.index), s); err != nil {
-			return err
+			return errors.New("redigo: error converting '" + string(name) + "' in ScanStruct: " + err.Error())
```
garyburd: Use fmt.Errorf.

garyburd:
How does the added information in the ScanSlice errors help an application?

connor4312:
Go doesn't build stack traces into errors. Even though it's often possible to logically follow where errors come from, I find that tagging errors in the "module" where they're created makes it much easier to track. For example, if there's something in a Redis list that we think will be a number, but isn't, then

```
redigo: error converting slice value: strconv.ParseInt: parsing \"\": invalid syntax
```

is much easier to track see than

```
strconv.ParseInt: parsing \"\": invalid syntax
```

in an application that might use multiple ParseInts.

PR #143, opened 2015-08-21
Handling SHUTDOWN command correctly

danielalves:
Currently, the `SHUTDOWN` command always returns an error. This PR fixes that.

### DIFF
--- a/internal/redistest/testdb.go
+++ b/internal/redistest/testdb.go
@@ -39,9 +39,16 @@ func (t testConn) Close() error {
 }
 
 // Dial dials the local Redis server and selects database 9. To prevent
-// stomping on real data, DialTestDB fails if database 9 contains data. The
+// stomping on real data, Dial fails if database 9 contains data. The
 // returned connection flushes database 9 on close.
 func Dial() (redis.Conn, error) {
+	return DialAndCheckDBSize(true)
+}
+
+// DialAndCheckDBSize dials the local Redis server and selects database 9. If
+// checkDBSize is true, the method fails if database 9 contains data. If it's
+// false, that check is skipped. The returned connection flushes database 9 on close.
+func DialAndCheckDBSize(checkDBSize bool) (redis.Conn, error) {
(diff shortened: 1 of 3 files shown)


garyburd:
io.EOF is the expected return from SHUTDOWN.  

     _, err := c.Do("SHUTDOWN")
     if err != io.EOF {
         // handle error
    }

danielalves:
@garyburd I see, but this is strange since `io.EOF` is the same error returned when a connection goes down, so I'm not able to know if everything went alright with the `SHUTDOWN` command or if something went wrong with the connection. Besides that, this is not really an *error*, this is the expected behavior (to close the connection after the shutdown), so I thought it would be better to return `nil, nil`. What do you think?

garyburd:
It's impossible to tell if shutdown executed correctly, but you do know that something went wrong if the returned error is not io.EOF.

I don't think it's worth adding a string allocation to the code path for this case.  Applications should either ignore the error or check for io.EOF.  Because documentation states that the command does not return a response, the application can also use this code:

     c.Send("SHUTDOWN")
    if err := c.Flush(); err != nil {
        // handle error
    }

danielalves:
@garyburd You can't tell a `SHUTDOWN` executed correctly, but you can tell it was received correctly (after all, we use a TCP connection). My point is `writeCommand` does not return an error, neither does `Flush`. Since we do not expect a response from a `SHUTDOWN`, it feels strange to return an error because of a `readReply` operation - EOF is associated with reading, but I'm not expecting to read anything with a `SHUTDOWN`.

garyburd:
`Do` is documented to write the command and read the reply. Read on a network connection closed by the peer is expected to return `io.EOF`. The SHUTDOWN command is documented to close the connection.  The expected error return is therefore `io.EOF`.

Call `Send` and `Flush` instead of `Do` if you do not want the client to read the reply that will not come.

Because the server can close the connection while the SHUTDOWN command is in flight between the client application and server application, it's impossible to tell if the command was received.

PR #145, opened 2015-08-30
Add DialURL to allow connection using a Redis URL

shosti:
After thinking about it some more, I think it's cleaner to take a URL
object instead of an unparsed string, but I think either way would work.

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -97,6 +97,15 @@ func DialConnectTimeout(d time.Duration) DialOption {
 	}}
 }
 
+// DialNetDial specifies a custom dial function for creating TCP
+// connections. If this option is left out, then net.Dial is
+// used. DialNetDial overrides DialConnectTimeout.
+func DialNetDial(dial func(network, addr string) (net.Conn, error)) DialOption {
+	return DialOption{func(do *dialOptions) {
+		do.dial = dial
+	}}
+}
+
 // DialDatabase specifies the database to select when dialing a connection.
 func DialDatabase(db int) DialOption {
 	return DialOption{func(do *dialOptions) {
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/url"
+	"regexp"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


redis/conn.go
```diff
@@ -104,6 +116,14 @@ func DialDatabase(db int) DialOption {
 	}}
 }
 
+// DialPassword specifies the password to use when connecting to
+// redis.
```
garyburd: "when connecting to the Redis server."

redis/conn.go
```diff
@@ -143,6 +163,55 @@ func Dial(network, address string, options ...DialOption) (Conn, error) {
...
+// scheme (https://www.iana.org/assignments/uri-schemes/prov/redis).
+func DialURL(u *url.URL, options ...DialOption) (Conn, error) {
+	if u.Scheme != "redis" {
+		return nil, fmt.Errorf("invalid redis URL scheme: %s", u.Scheme)
+	}
+
+	// As per the IANA draft spec, the host defaults to localhost and
+	// the port defaults to 6379
+	host, port, err := net.SplitHostPort(u.Host)
+	if err != nil {
```
garyburd: Parsing error text is fragile. Assume that the port is not specified if err != nil.  

    if err != nil {
        host = u.Host
        port = "6379"
    } else if host == "" {
        host = "localhost"
    }

redis/conn_test.go
```diff
@@ -422,6 +424,105 @@ func TestReadDeadline(t *testing.T) {
...
+func TestDialURL(t *testing.T) {
+	var (
+		u   *url.URL
+		err error
+	)
+	for _, d := range dialErrors {
+		u, _ = url.Parse(d.rawurl)
+		_, err = redis.DialURL(u)
+		if err == nil || !strings.Contains(err.Error(), d.expectedError) {
+			t.Errorf("NewDialer did not throw correct error (expected %v to contain %s)", err, d.expectedError)
```
garyburd: Replace "throw" with "return".  Errors are not thrown in Go.

garyburd:
Use a string to specify the URL.  It's simpler for applications.  This is not a scenario where *url.URL should be used to avoid the design flaw in the url.URL type.

To avoid destroying user data when running tests, check to see if the database is empty before writing to the database.  Is there a way to check the selected database without writing to the database?

See inline comments on code.

redis/conn.go
shosti: I did originally, but there are some actual parse errors that are useful to return (for instance, if there are too many colons). I'll change it back, though, I guess invalid hosts will eventually be caught trying to connect.

shosti:
OK, I updated the pull request. Unfortunately, I don't think there's any way to check which database is selected without writing to the database (correct me if I'm wrong), so I just added a check to make sure that DBSIZE is 0. Let me know if you notice anything else.

PR #147, opened 2015-09-01
Support For Google App Engine 

Karthic-Hackintosh:
I have done modifications to allow support for Google App Engine . TCP connection and Dialer implementation is done using appengine/socket . Since bulid flags are used appengine version for the connection ( conn_appengine.go ) is ignored during normal build , but considered only during the AppEngine build


garyburd:
No modifications should be required after #145 lands.  You should be able to:

    c, err := redis.Dial(network, host, redis.DialNetDial(func(network, addr string) {
          return socket.Dial(appEngineContext, network, address) }))

PR #149, opened 2015-09-01
Fix bux causing Ints call to swallow error returned from the connecton object

szank:
Like in the description. 

Discovered it when i stopped redis server, but my app was returning errNil and not connection errors


garyburd:
Thank you for finding this and the fix.

PR #153, opened 2015-09-21
Added support for "omitempty" tag

kpumuk:
In my current project I need to skip empty struct fields. I noticed you have started working on "omitempty" tag, so I went ahead and implemented it.

### DIFF
--- a/redis/scan.go
+++ b/redis/scan.go
@@ -523,11 +523,22 @@ func flattenStruct(args Args, v reflect.Value) Args {
 	for _, fs := range ss.l {
 		fv := v.FieldByIndex(fs.index)
 		if fs.omitEmpty {
-			if (fv.Kind() == reflect.Map || fv.Kind() == reflect.Slice) && fv.Len() == 0 {
-				continue
+			var empty = false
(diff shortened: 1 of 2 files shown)


garyburd:
Use this code to check for empty:

	switch v.Kind() {
	case reflect.Array, reflect.Map, reflect.Slice, reflect.String:
		return v.Len() == 0
	case reflect.Bool:
		return !v.Bool()
	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
		return v.Int() == 0
	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
		return v.Uint() == 0
	case reflect.Float32, reflect.Float64:
		return v.Float() == 0
	case reflect.Interface, reflect.Ptr:
		return v.IsNil()
	}

kpumuk:
Updated pull request

garyburd:
I squashed the commits to 7a0cba7bcb6e1ea22c7d2d4b082d804a7911b319

PR #176, opened 2016-03-11
Add custom argument representation handling

stevenh:
Add the ability for types to control how they are sent when used as arguments to redis commands.

This is useful for types which don't want to use their Stringer representation when used as an argument to a redis command.

Examples of this is a time.Duration which the user wants to store the number of seconds or a time.Time being stored as int instead of a string.

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -322,6 +322,10 @@ func (c *conn) writeCommand(cmd string, args []interface{}) (err error) {
 			}
 		case nil:
 			err = c.writeString("")
+		case Argument:
+			var buf bytes.Buffer
+			fmt.Fprint(&buf, arg.RedisArg())
+			err = c.writeBytes(buf.Bytes())
 		default:
 			var buf bytes.Buffer
 			fmt.Fprint(&buf, arg)
--- a/redis/conn_test.go
+++ b/redis/conn_test.go
@@ -459,7 +459,7 @@ var dialErrors = []struct {
 		"localhost",
 		"invalid redis URL scheme",
 	},
-	// The error message for invalid hosts is diffferent in different
+	// The error message for invalid hosts is different in different
 	// versions of Go, so just check that there is an error message.
 	{
 		"redis://weird url",
--- a/redis/conn_test.go
+++ b/redis/conn_test.go
@@ -46,6 +46,14 @@ func dialTestConn(r io.Reader, w io.Writer) redis.DialOption {
 	})
 }
 
+type durationArg struct {
+	time.Duration
+}
+
+func (t durationArg) RedisArg() interface{} {
+	return t.Seconds()
+}
+
 var writeTests = []struct {
 	args     []interface{}
 	expected string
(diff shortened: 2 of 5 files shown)


garyburd:
The application should convert arguments to an appropriate type before calling Do or Send.

It was a mistake that the connection used Stringer.

stevenh:
Having to convert everything manually before passing it doesn't make for clean flow.

Allowing types to control their redis conversion has been very useful here.

The core database/sql uses a similar technique, too so we think this would be a nice addition, if you would reconsider?

garyburd:
> Examples of this is a time.Duration which the user wants to store the number of seconds or a time.Time being stored as int instead of a string.

A wrapper type is required to adapt time.Duration to the interface in this PR.  The use of the wrapper type does not seem better than calling d.Seconds().

stevenh:
Yes that is the case so possibly a bad example, in this case we already had a wrapper class for time.Duration, for consistent API serialisation to and from JSON so its was just an extra method.

We also use this for pointer types (supporting nil), times which we want stored in milliseconds.

Possibly the most useful is for integer based types used for enums e.g.
```go
// VirtualTypeID represents a machines virtual type
type VirtualTypeID int

// Machine virtual types
const (
        _                                 = iota
        VirtualTypePhysical VirtualTypeID = iota
        VirtualTypeJail
        VirtualTypeCloud
)

// RedisArg satisfies redis.Argument interface
func (vt VirtualTypeID) RedisArg() interface{} {
        return int(vt)
}
```
The key thing is if you do have a type which needs special treatment, you define RedisArg once on the type and you're done, you don't have to remember to convert them all the time, just pass them in and the right thing happens :)

garyburd:
The PR does not change the encoding for values of type `VirtualTypeID`.  WIth or without the PR,  the values are formatted as decimal text.

stevenh:
Sorry missing an important bit from that
```go
func (vt VirtualTypeID) String() string {
        switch vt {
        case VirtualTypePhysical:
                return "Physical"
        case VirtualTypeJail:
                return "Jail"
        case VirtualTypeCloud:
                return "Cloud"
        default:
                return fmt.Sprintf("Unknown (%d)", vt)
        }
}
```

jlburkhead:
Should add this to scanning as well

garyburd:
The PR should include scan as suggested by @jlburkhead.

stevenh:
Indeed that's a good idea, I'll look to get the PR updated soon.

stevenh:
@garyburd sorry this took a while fell off my radar.

@jlburkhead thanks for the initial patch. I had to change a few things to get it to work, so would be good if you could test for your use case. Also your changes to writeCommand didn't seem needed, if you have a use case which requires them please let me know so I can add a test for it.

stevenh:
@garyburd just checking if you're happy with this now or would like any further changes?

garyburd:
I will look at this this week.  Thank you for reminding me.

stevenh:
@garyburd just checking in, in case it slipped off your radar ;-)

garyburd:
Thank you for being patient.

Issue #99, opened 2015-01-04
TLS support


uhoh-itsmaciek:
There's some movement to [add TLS support to mainline Redis](https://github.com/antirez/redis/issues/2178) now. Since this basically just layers the existing protocol on top of TLS (e.g., see [a Ruby client patch](https://github.com/redis/redis-rb/pull/419/files#diff-20ca4377c0eb8c090a4e21893b08e8f9R216)), I think one can already use redigo with this by passing in a custom dialer, but it might be nice to have better out-of-the-box support. Any thoughts on whether this is a good idea, and if so, what it should look like? Perhaps a `redis.DialTLS`?

garyburd:
I'll add the feature soon after antirez commits the change. The API will look like http.Transport.

uhoh-itsmaciek:
Awesome! Thank you.

garyburd:
Fixed in #186.

PR #186, opened 2016-06-01
Add a DialOption for passing a tls.Config, use tls dialer if config is

kenkouot:
This PR adds support for dialing a tcp connection over TLS to redigo. The main use case for this is *TLS offloading* using something like HAProxy and allows for the redigo client to talk to a redis that is securely proxied.

I saw that #99 is still open, but feel like while this PR and that issue are related, they are ultimately orthogonal. This adds support for people using proxy-based solutions like HAProxy, or perhaps other solutions like stunnel. When antirez adds mainline TLS, this feature might be better served by native TLS support in Redis (or Sentinel), but for now it's fairly common to have HAProxy in front of a Sentinel setup.

ps thanks for the great work on redigo!

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -17,6 +17,7 @@ package redis
 import (
 	"bufio"
 	"bytes"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


garyburd:
The package provides the [DialNetDial option](https://godoc.org/github.com/garyburd/redigo/redis#DialNetDial) option for this scenario.

     o := redis.DialNetDial(func(network, addr string) (net.Conn, error) {
              return tls.DialWithDialer(&net.Dialer{Timeout: 1 * time.Second,}, network, addr, tlsConfig)
      })

kenkouot:
I should have read the documentation more carefully :D thanks!

PR #195, opened 2016-08-04
Adding ConnMutex for threadsafe Do()

tbarron-xyz:
`ConnMutex` is a wrapper for `Conn` where its `Do` method is wrapped in a `sync.Mutex`. It only implements the `Do` method. `redis.DialMutex` has the same syntax as `redis.Dial` and returns a `ConnMutex`. The `Do` method has the same syntax as for the `Conn` interface.

```
client := redis.DialMutex("tcp", ":6379")
reply, err := client.Do("INCR", "myvalue", 1)
```

(My application wants a lot of concurrent `Do()` calls, and I got tired of exhausting a pool under heavy load, so I started using a single connection and wrapping everything in a `sync.Mutex`. I thought this may be useful to more than just myself.)

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -568,3 +568,24 @@ func (c *conn) Do(cmd string, args ...interface{}) (interface{}, error) {
 	}
 	return reply, err
 }
+
+type connMutex struct {
+	Conn
+	Mutex *sync.Mutex
+}
+
+func (c *connMutex) Do(cmd string, args ...interface{}) (interface{}, error) {
+	c.Mutex.Lock()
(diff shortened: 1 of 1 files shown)


garyburd:
If this approach is acceptable for your application, then you can avoid exhausting the pool by setting [Pool.Wait](https://godoc.org/github.com/garyburd/redigo/redis#Pool.Wait) to true.

I do not want to take this change because there are better approaches for sharing a single connection.  One example is [redisx.ConnMux](https://godoc.org/github.com/garyburd/redigo/redisx#ConnMux).

PR #210, opened 2016-11-28
Treat named primitive types as primitives, rather than as strings.

JaredReisinger:
As an interim measure to a more full-featured argument serializer interface, handle simply-wrapped types as their primitives rather than falling back to Stringer-style conversion.

This is particularly useful in the case of enum/iota-style constants that are typed with a specific name.  Previously, any `String()` method on this type would result in those stringified values getting written to redis, and then a parse failure when read, especially when read via `ScanStruct()`, where it's difficult to intervene in the parsing.  With this change, those values are written as their primitive (numeric) type, and thus read as that primitive type, and can be successfully assigned to the named type even when it's a member of a struct.

As an example:

```go
type status int

const (
    StatusUnknown status = iota,
    StatusBad,
    StatusGood,
)

func (s status) String() string {
    switch s {
    case StatusBad:
        return "bad"
    case StatusGood:
        return "good"
    default:
        return "????"
    }
}

struct Data {
    Status status `redis:"status"`
    // ... other fields ...
}

func writeData(conn redis.Conn, hashKey string, data *Data) err {
    _, err := conn.Do("HMSET", redis.Args{}.Add(hashKey).AddFlat(data)...)
    return err
}

func getData(conn redis.Conn, hashKey string) *Data, error {
    data := &Data{}
    err := redis.ScanStruct(redis.Values(conn.Do("HGETALL", hashKey)), data)
    return data, err
}
```

Before this change, redis would have `status: "bad"` (or `status: "good"`) written , and `ScanStruct()` would later fail to parse that as an int.  _With_ this change, redis will get `status: 1` (or `status: 2`) written, and the later `ScanStruct()` will successfully parse the number and assign it to the `Status` member.

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net"
 	"net/url"
+	"reflect"
 	"regexp"
 	"strconv"
 	"sync"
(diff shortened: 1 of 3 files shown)


redis/conn.go
```diff
@@ -351,32 +352,50 @@ func (c *conn) writeCommand(cmd string, args []interface{}) (err error) {
...
+		if arg {
+			err = c.writeString("1")
+		} else {
+			err = c.writeString("0")
+		}
+	case nil:
+		err = c.writeString("")
+	default:
+		v := reflect.ValueOf(arg)
+		switch v.Kind() {
```
JaredReisinger: I'm unsure of the potential performance penalty to `reflect.ValueOf()`... if it's inexpensive, it may make more sense to _always_ do this, and use the `v.Kind()` switch on line 361 instead of the `arg.(type)` switch.  That would avoid the recursion and make this a little less complicated-looking.

garyburd:
> As an interim measure to a more full-featured argument serializer interface

What is the full-featured argument serializer interface that you are referring to?

garyburd:
This change will break applications that rely on the fallback to Stringer.

PR #217, opened 2016-12-10
Fixes AUTH error in example newPool function

### LINKED ISSUE #216: Example newPool function throw error when supplied with empty password during AUTH
I used the example newPool function to create a new connection pool. When I used it I got the error `ERR Client sent AUTH, but no password is set`. This happens as a result of the execution of the `AUTH` command with the empty password as parameter.

I did fix the error in my version. The function should be
```go
func newPool(server, password string) *redis.Pool {
  return &redis.Pool{
    MaxIdle:     3,
    IdleTimeout: 240 * time.Second,
    Dial: func() (redis.Conn, error) {
      c, err := redis.Dial("tcp", server)
      if err != nil {
        return nil, err
      }
      if password != "" {
        if _, err = c.Do("AUTH", password); err != nil {
          c.Close()
          return nil, err
        }
      }
      return c, err
    },
    TestOnBorrow: func(c redis.Conn, t time.Time) error {
      if time.Since(t) < time.Minute {
        return nil
      }
      _, err := c.Do("PING")
      return err
    },
  }
}
```

GamingCoder:
This change fixes #216

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -58,9 +58,11 @@ var (
 //              if err != nil {
 //                  return nil, err
 //              }
-//              if _, err := c.Do("AUTH", password); err != nil {
-//                  c.Close()
-//                  return nil, err
+//              if password != "" {
+//                  if _, err := c.Do("AUTH", password); err != nil {
+//                      c.Close()
+//                      return nil, err
+//                  }
 //              }
 //              return c, err
 //          },


garyburd:
See my edited comment at https://github.com/garyburd/redigo/issues/216#issuecomment-266230242

PR #237, opened 2017-03-06
support LimitedReader as conn.Send arg

GlenDC:
Currently it is not possible to set a value directly from a reader, which means that one would first have to write to an intermediate buffer (eg a byte slice) and use that buffer then to SET the value.

This PR makes it possible to directly pass a `*io.LimitedReader` as a arg to the `conn.Send` function, making it possible to pipe that reader effectively directly into the internal `Writer`.

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -344,6 +344,32 @@ func (c *conn) writeFloat64(n float64) error {
 	return c.writeBytes(strconv.AppendFloat(c.numScratch[:0], n, 'g', -1, 64))
 }
 
+func (c *conn) writeFromLimitedReader(lr *io.LimitedReader) error {
+	if lr == nil || lr.R == nil || lr.N == 0 { // ensure there is data to be read
+		return protocolError("no data to be read from LimitedReader")
+	}
+
+	// write fixed length
+	fixedLength := lr.N
+	c.writeLen('$', int(fixedLength))
+
+	// write actual content, capped at limitedReader
+	n, err := c.bw.ReadFrom(lr)
+	if err != nil {
+		return err
+	}
+
+	// ensure that written length is as much as the fixed length we specified earlier
+	if n != fixedLength {
+		return protocolError(fmt.Sprintf(
+			"expected to read %d bytes from limitedReader, while only read %d",
+			fixedLength, n))
+	}
+
+	_, err = c.bw.WriteString("\r\n")
+	return err
+}
+
 func (c *conn) writeCommand(cmd string, args []interface{}) (err error) {
 	c.writeLen('*', 1+len(args))
 	err = c.writeString(cmd)
@@ -370,6 +396,8 @@ func (c *conn) writeCommand(cmd string, args []interface{}) (err error) {
 			}
 		case nil:
 			err = c.writeString("")
+		case *io.LimitedReader:
+			err = c.writeFromLimitedReader(arg)
 		default:
 			var buf bytes.Buffer
 			fmt.Fprint(&buf, arg)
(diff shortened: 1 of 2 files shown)


GlenDC:
@garyburd this is my first contribution to this Repo, so please do let me know if I forgot something obvious. I hope we can make this into a feature that gets merged into master. If it requires extra work, please do let me know as well.

garyburd:
Let's talk about the problem you are trying to solve.  What kind of io.Readers is your application working with? Is it important to limit the readers or are you using LimitedReader to bundle a length and io.Reader as a single argument?

GlenDC:
Thanks for the quick response @garyburd!

First I was adding a new Type that bundles the length with a reader, but then I realized that `io` already has type for that, called `LimitedReader`. So that's why I thought that it might be useful to use that one, as it also allows people to use a `io.Reader` object that might have more content than the length they specify. However as the `io.Reader` might have less content, I make sure to check that the read length is the same as the max Length, as I expect that Redis otherwise might act quite weird?

But in short what I am trying to solve is that I can pipe my Reader directly into Redis, as it would be silly to first have to read into an intermediate buffer. My Reader is also a `net/Conn` so the length is something I specify, so I can write to Redis in blocks of a fixed length. Our previous API was using byte slices, and reading it into intermediate byte slices, so we were just writing byte slices, but now our API is cleaner and better, and using `io.Reader`, which is when I realized that for now it didn't seem to be possible to pipe that reader directly into Redis, hence this PR.

garyburd:
Is your application working with a *net.TCPConn or some other type?

GlenDC:
Yes it is, at least for some cases, might be for all, but I'm not sure. The connection is passed in to my backend as a `io.Reader` here: https://github.com/g8os/gonbdserver/blob/master/nbd/connection.go#L324; the Backend that makes use of your Redis client is in some other repository. But you can imagine that I'm giving that reader straight to the forked version of your Redis client.

GlenDC:
Oops sorry, closed it by accident. Wrong button...

garyburd:
OK, I understand the problem that you are trying to solve and I am now looking at the code.

If `writeFromLimitedReader` does not write exactly `fixedLength` bytes to `c.bw`, then [`c.fatal`](https://github.com/GlenDC/redigo/blob/cf2ab19f764ae4555be05ac4f5d46fd73aa6f0ef/redis/conn.go#L289) must be called.  The Redigo connection is not useable after fatal is called.  The feature is fragile.

Does this feature significantly improve the performance of your application? If it's just to tidy up your code, then I do not want to take this change.  The cost of supporting a fragile feature is not worth saving your application a few lines of code.

GlenDC:
It does save performance significantly, as all its supposed to do is read and write, so if that server would have to do that first via an intermediate buffer, than there is quite some overhead. In my case that length is guaranteed to be correct. But I do see your point.

If you want we could make a custom Reader, that does not only limit the Reading, but that also starts writing `0`'s for all the missing content (in case the available content is smaller than the specified length). That would keep the Redigo connection stable no, @garyburd ?

garyburd:
Did you measure the performance?  It's not obvious that that this change will reduce the amount of copying.  The bufio.Writer type writes directly to the underlying writer for large writes and buffers small writes.  This change may be trading copying to your intermediate buffer for copying in bufio.Writer. It's also not obvious that copying data is a significant cost.

Because your application reads from a network connection, there's no guarantee that the specified number of bytes will be read.

Silently filling in missing data is not a good idea.

GlenDC:
I agree that silently filling missing data isn't a good idea, but it's the only idea I can come up for now to keep the connection stable.

I'm performing some benchmarks later today. I will share the results with you, and we can take it from there. Thank you for the constructive feedback, I appreciate that a lot.

GlenDC:
After some continued testing, profiling and experimenting, it seems like you were right indeed. Closing this PR as I don't think this proposition is word the trouble we would get as a result. Thanks for the feedback and input!

PR #245, opened 2017-04-18
Implement geopos reply helper function

### LINKED ISSUE #230: Add reply helper functions for GEO commands
Add [reply helper functions](https://godoc.org/github.com/garyburd/redigo/redis#hdr-Reply_Helpers) for the GEO commands. 

Note: Redigo supports all GEO commands. This issue is about adding convenience functions for working with GEO command results.

crackerplace:
Implemented a helper function for geopos reply.

@garyburd 

Fixes #230

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -391,3 +391,35 @@ func Int64Map(result interface{}, err error) (map[string]int64, error) {
 	}
 	return m, nil
 }
+
+// Positions is a helper that converts an array of positions (lat, long)
+// into a [][2]float64. The GEOPOS command returns replies in this format.
+func Positions(result interface{}, err error) ([]*[2]float64, error) {
+	values, err := Values(result, err)
+	if err != nil {
+		return nil, err
+	}
+	positions := make([]*[2]float64, len(values))
+	for i := range values {
+		if values[i] == nil {
+			continue
+		}
+		p, ok := values[i].([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("redigo: unexpected element type for Interface slice, got type %T", values[i])
+		}
+		if len(p) != 2 {
+			return nil, fmt.Errorf("redigo: unexpected number of values for a member position,got %d", len(p))
+		}
+		lat, err := Float64(p[0], nil)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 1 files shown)


garyburd:
The function should handle null values without returning an error.  One way is to return a slice of pointers to lat/long pairs:

     func Positions(result interface{}, err error) ([]*[2]float64, error)

crackerplace:
@garyburd done.

redis/reply.go
```diff
@@ -394,13 +394,16 @@ func Int64Map(result interface{}, err error) (map[string]int64, error) {
...
 		return nil, err
 	}
-	positions := make([][2]float64, len(values))
+	positions := make([]*[2]float64, len(values))
 	for i := range values {
+		if values[i] == nil {
+			continue
+		}
 		p, ok := values[i].([]interface{})
 		if !ok {
```
garyburd: Also check len(p) >= 2

crackerplace: If the position for a given input member is returned in the response,do we have to even check if the position contains both latitude and longitude ? Does geoadd command allow not specifying either latitude or longitude for a given member.

garyburd: Yes. The function should return an error in the case of an unexpected response. The function should not panic.

crackerplace: ok.In that case,I will return error if if len(p) != 2.

crackerplace:
@garyburd Handled the length of the position array.

redis/reply.go
```diff
@@ -391,3 +391,35 @@ func Int64Map(result interface{}, err error) (map[string]int64, error) {
...
+	positions := make([]*[2]float64, len(values))
+	for i := range values {
+		if values[i] == nil {
+			continue
+		}
+		p, ok := values[i].([]interface{})
+		if len(p) != 2 {
+			return nil, fmt.Errorf("redigo: unexpected number of values for a member position,got %d", len(p))
+		}
+		if !ok {
```
garyburd: The true branch of this if statement will never be executed because len(p) != 2 if ok == false.

crackerplace: @garyburd yes,I moved it to the right position.thanks.

crackerplace:
@garyburd squashed the commits.thanks.

PR #246, opened 2017-04-27
Add Int64s helper function

### LINKED ISSUE #192: add a Int64s() function
AFAIK there is no function Int64s for redis reply, I thinks it is necessary for a int64 array.

crackerplace:
Implemented a helper function for int64s reply.

@garyburd

Fixes #192

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -317,6 +317,20 @@ func Ints(reply interface{}, err error) ([]int, error) {
 	return ints, nil
 }
 
+// Int64s is a helper that converts an array command reply to a []int64. If
+// err is not equal to nil, then Int64s returns nil, err.
+func Int64s(reply interface{}, err error) ([]int64, error) {
+	var int64s []int64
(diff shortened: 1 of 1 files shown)


garyburd:
Call Int64 in a loop to avoid use of reflection in ScanSlice.

crackerplace:
@garyburd Sure.If needed I can modify Ints function as well to avoid using ScanSlice.

garyburd:
Yes, that will be helpful

crackerplace:
Sorry,some delay in this.Will be done by weekend.

crackerplace:
@garyburd How should we handle nil in the reply interface slice.Should we ignore and continue and in which case the specific entry in the result slice will be the zero value.ScanSlice has similar logic.

garyburd:
Yes, use the zero value.  

If an application should call Values directly if it needs to distinguish between nil and zero in the result.

PR #247, opened 2017-05-10
Scan skip string when dest is nil

bits01:


bits01:
Not sure what it failed for Go 1.4 doesn't look related to my change.

PR #259, opened 2017-06-30
Update README to add the related projects xgfone/go-redis

xgfone:
I implement a high-level API interface based on Redigo.


garyburd:
No thanks. I do not want to recommend a package that handles errors by panicking.

PR #261, opened 2017-07-04
Update scan.go

smartwalle:
Update ScanStruct method


garyburd:
Explain the purpose of this change.

PR #288, opened 2017-10-30
Added support for Floats slices

mshtmfv:


redis/scan.go
```diff
@@ -134,6 +150,8 @@ func convertAssignValue(d reflect.Value, s interface{}) (err error) {
 		err = convertAssignBulkString(d, s)
 	case int64:
 		err = convertAssignInt(d, s)
+	case float64:
```
garyburd: Explain why this is needed.  

Redigo does not use floats to represent Redis values.

redis/reply.go
```diff
@@ -317,6 +317,22 @@ func Ints(reply interface{}, err error) ([]int, error) {
...
+
+// Floats is a helper that converts an array command reply to a []float64. If
+// err is not equal to nil, then Floats returns nil, err.
+func Floats(reply interface{}, err error) ([]float64, error) {
+	var floats []float64
+	values, err := Values(reply, err)
+	if err != nil {
+		return floats, err
+	}
+	if err := ScanSlice(values, &floats); err != nil {
```
garyburd: Convert the values directly instead of using ScanSlice.

mshtmfv: Why I can't use ScanSlice?

redis/scan.go
mshtmfv: I need to store float lists in Redis so I need to get them back somehow to golang client.

redis/reply.go
garyburd: Write it directly because the code is easier to understand and faster than ScanSlice.

mshtmfv: Ok, I would change code soon.

redis/scan.go
garyburd: The Redis result types are error, simple string, bulk string, integer and array.  Redigo does not use float to represent any of these types.

PR #294, opened 2017-11-21
Enable keep-alive on server connections

stevenh:
Add the option to configure connection keep-alive using the new DialKeepAlive option.

This is set to 5 minutes by default to ensure that half-closed connections are detected.

Without this Pub/Sub connections can get disconnected by the redis server and we will never notice which can potentially deadlock applications.

Also:
* Add --redis-address <address> command line option to tests defaulting to 127.0.0.1. Without this all tests fail on machines with multiple addresses.


stevenh:
This fixes #295

PR #297, opened 2017-11-22
Disable read timeout for Pub/Sub connections

stevenh:
Due to the nature Pub/Sub connections they highly likely to block on reads for long periods hence it doesn't make sense to apply any read timeout configured on the pool to Pub/Sub connections.

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -82,6 +82,7 @@ type dialOptions struct {
 }
 
 // DialReadTimeout specifies the timeout for reading a single command reply.
+// Read timeout is not used by Pub/Sub connections.
 func DialReadTimeout(d time.Duration) DialOption {
 	return DialOption{func(do *dialOptions) {
 		do.readTimeout = d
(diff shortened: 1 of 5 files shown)


stevenh:
For reference the alternative to this would be to test errors return by PubSubConn.Receive() in all consumers to check for net.Error.Timeout() == true and ignore them, which will possible isn't very idiomatic.

garyburd:
I am rejecting this PR because this is a breaking change.  Also,  ReadWithoutTimeout is not useful in a robust application. 

PING with a duration shorter than the [read timeout](https://godoc.org/github.com/garyburd/redigo/redis#DialReadTimeout).

stevenh:
I suspected this may be the case, something else to try an solve with the [high level API](https://github.com/garyburd/redigo/issues/296)

We'll keep this internally so we don't have to update all consumers until the new Pub/Sub API is available.

garyburd:
See #298

PR #311, opened 2018-01-18
add Uint64s  reply function

edwardhey:
FYI.

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -345,6 +345,24 @@ func Int64s(reply interface{}, err error) ([]int64, error) {
 	return result, err
 }
 
+func Uint64s(reply interface{}, err error) ([]uint64, error) {
+	var result []uint64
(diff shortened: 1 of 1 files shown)


redis/reply.go
```diff
@@ -345,6 +345,24 @@ func Int64s(reply interface{}, err error) ([]int64, error) {
 	return result, err
 }
 
+func Uint64s(reply interface{}, err error) ([]uint64, error) {
+	var result []uint64
+	err = sliceHelper(reply, err, "Uint64s", func(n int) { result = make([]uint64, n) }, func(i int, v interface{}) error {
+		switch v := v.(type) {
+		case uint64:
```
garyburd: This case will not be executed.  Add int64 case and return error on negative number.

garyburd:
Let me know if you plan to address my comments or if I should close this PR.

Issue #296, opened 2017-11-22
Add high-level pubsub API


garyburd:
Add high-level pubsub API that accomplishes the following goals:

- Hides receive loop from application, possibly by calling application supplied function to handle notifications.
- Uses context.Context to cancel subscriptions.
- Provides way for application to backfill after dialing and subscribing.
- Uses ping to check for live connection and server.
- Optional: support using pool to get connections.
- Optional: subscribe/unsubscribe after running.

rubenv:
For those looking at something like this: https://github.com/rubenv/rrpubsub

Wrote this a long time ago (still works great!) and ticks pretty much all these boxes.

PR #336, opened 2018-04-27
Spike for SubscriptionReader

nelz9999:
Related to #296 

This is potential helper functionality for using PubSub. (I did this as a PR instead of an Issue, just so folks could see what I was talking about.)

I expect the responses may include:
* "Great, but it's overly specific/opinionated..."
  * "... so maybe make it more generic or more configurable."
  * "... so we don't want it in the main project, but maybe in a `helper`/`examples` package."
* "Looks good, but need docs and the unit tests." (Happy to do! Just wanted some direction before investing more time.)

Whaddya think?

### DIFF
--- a/redis/reader.go
+++ b/redis/reader.go
@@ -1,0 +1,102 @@
+// Copyright 2012 Gary Burd
+//
+// Licensed under the Apache License, Version 2.0 (the "License"): you may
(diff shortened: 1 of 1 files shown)


nelz9999:
(I just realized I screwed up the semantics of `WriteTo`, which I'll fix in the full-er checkin.)

PR #340, opened 2018-05-19
Add a stub object for testing redis.Conn

nelz9999:
I frequently find myself wanting a helper for testing that I got an abstract `redis.Conn`. (Especially when trying to  abstract away the distinctions between `redigo` and `redisc`.)

I thought others might also benefit from this work.

### DIFF
--- a/stubs/conn.go
+++ b/stubs/conn.go
@@ -14,14 +14,14 @@
 
 package stubs
 
-import "fmt"
+import "errors"
 
-// MsgNotImplemented is the default error body returned when a
-// method is invokes without a stub func defined.
-const MsgNotImplemented = "not implemented"
+// ErrNotImplemented is the default error returned when a
+// method is invoked without a stub func defined.
+var ErrNotImplemented = errors.New("stub: not implemented")
 
 // StubConn is a redis.Conn that helps consumers of redigo with testing
-type StubConn struct {
+type Conn struct {
 	OnClose   func() error
 	OnErr     func() error
 	OnDo      func(commandName string, args ...interface{}) (reply interface{}, err error)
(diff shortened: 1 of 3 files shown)


nelz9999:
Hrm... Looking at the build errors, it's for Go <= 1.6... I guess `T.Run(...)` was added in 1.7.

☹️ I'll try to find some time to circle back and fix the tests.

nelz9999:
Updated tests for pre-1.7 builds.

redmondnoodle:
A more idiomatic way to handle the error is:

    // ErrNotImplemented is the error returned when a method invokes a stub function with no callback.
    var ErrNotImplemented = errors.New("not implemented")

    ...

    return ErrNotImplemented

As currently written, the application must compare error strings to test for the not implemented error.  This test is fragile.

A more idiomatic name for the type is `Conn`.  The "Stub" part of the name is stutter.  See https://blog.golang.org/package-names for comments about stutter. 

(Feel free to ignore this advice as I am not a package maintainer).

nelz9999:
Updated based on @redmondnoodle feedback.

NB: In my own projects, I've been trending away from the "sentinel error" pattern, partially because publicly mutable values sort of freak me out. But, I realize that it is on me to conform to standard idioms when contributing to a public project.

garyburd:
I don't get the sense that this is something the applications will use, but I'll leave this open in case the new Redigo maintainers think otherwise.

A more idiomatic approach is to provide a type that can be embedded in another struct:

```
type NotImplementedConn struct {}
func (NotImplementedConn)  Do(cmd string, args ...interface{}) (interface{}, error) { return nil, ErrNotIMplemented }
func (NotImplementedConn)  Send(cmd string, args ...interface{}) error { return ErrNotIMplemented }
func (NotImplementedConn) Receive() (interface{}, error) { return nil, ErrNotMplemented }
... and so on
```

nelz9999:
Whoa. 

I hadn't come across that before. And when I went to look it up, I found this suggestion: https://medium.com/stupid-gopher-tricks/stubbing-out-interfaces-in-go-e85afc200aa8

Basically, you just embed the interface into the concrete stub struct, and anything not implemented panics.

I see that this technique totally invalidates any need for a stub class like I submitted. Closing.

PR #344, opened 2018-05-29
Improve tests suite for nil pointer struct scans

darkliquid:
When scanning structs, the test suite wasn't properly verifying that
a pointer field should be nil if the redis field doesn't exist.


garyburd:
Reduce the new test case to the nil pointer only.

PR #363, opened 2018-08-30
Migrate to Go Modules

marwan-at-work:
This PR migrates the package to Go Modules and updates the path according to Semantic Import Versioning

Generated by https://github.com/marwan-at-work/mod

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,1 @@
+module github.com/gomodule/redigo/v2
--- a/internal/redistest/testdb.go
+++ b/internal/redistest/testdb.go
@@ -19,7 +19,7 @@ import (
 	"errors"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/v2/redis"
 )
 
 type testConn struct {
--- a/redis/conn_test.go
+++ b/redis/conn_test.go
@@ -28,7 +28,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/v2/redis"
 )
 
 type testConn struct {
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -25,7 +25,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/gomodule/redigo/internal"
+	"github.com/gomodule/redigo/v2/internal"
 )
 
 var (
--- a/redis/pool17_test.go
+++ b/redis/pool17_test.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"testing"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/v2/redis"
 )
 
 func TestWaitPoolGetContext(t *testing.T) {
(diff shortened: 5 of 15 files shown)


itamarhaber:
@marwan-at-work what's the reference/context?

(sorry, newbie in this project)

garyburd:
This PR breaks compatibility with Go 1.8 and below.

marwan-at-work:
@garyburd yes, totally up to you if you'd like to accept that or not :)

garyburd:
Is there a way to  add Go module support and retain backwards compatibility with older versions of Go?

marwan-at-work:
@garyburd go 1.9.X and 1.10.X (x being latest) are backwards compatible. They can still import your library without the vX suffix. But otherwise, 1.8 won't work unless you copy and paste the code into a vX directory (or type alias things there) which is not a fun thing to do.

elithrar:
Go 1.9 was released in August 2017 - only 1 year ago. Since modules are only "experimental" in Go 1.11, we could hold off until Go 1.12 (when they are first-class), which is scheduled for Jan 2019.

Packages that declared a >= v2.x on the "vanilla" import path are effectively incompatible with modules. I believe @natefinch ran into this with [mage](https://github.com/magefile/mage) and @markbates with Buffalo.

natefinch:
yeah, so I was actually wrong.... with the latest patch of 1.9 and 1.10, those tools understand how to read go.mod... so if your code is on v2, but your import path doesn't include v2.... the go tool will figure it out if you put 
```
module github.com/foo/bar/v2 
```
in the go.mod, and it'll use the v2 branch of the repo.  

At least, that's what I've been told by reliable sources.

natefinch:
(but also if you want to be compatible with go 1.8 or earlier, you need a physical /v2 folder, because those tools weren't updated to read the go.mod and don't know)

What I'd do: try it out before merging.

garyburd:
What are the downsides of not migrating to Go Modules?

marwan-at-work:
@garyburd for redigo not much. 

The upside from introducing Go Modules is that the Go Command now recognizes your dependencies and can resolve them appropriately when users use redigo in their own systems. But since redigo does not have any dependencies it seems like, then you're fine. 

Another one, is that it's just a migration thing. Which can wait until 1.12 :)

garyburd:
Importers using Go 1.8 and below will vote against this PR.

Will this work?

- Remove redigo/internal by copying what's needed to  redigo/redis and redigo/redisx.
- Remove the import comment from the package declaration.
- Update _test.go and redigo/redisx files to import github.com/gomodulue/redigo/v2/redis instead of github.com/gomodule/redis. Use build tags to disable these files on Go 1.8 and below.
- Add the go.mod file.

With this change, redisx and some tests are lost on Go 1.8 and below, but redigo/redis will continue to work.

marwan-at-work:
@garyburd as another option: is it acceptable for users of 1.8 and below to simply stick to the previous major version of redigo until they are ready to migrate Go 1.8 to 1.9 or above?

garyburd:
It is not acceptable to require users to change how they fetch the code for this package.  This includes Go 1.8 users who fetch the package using `go get`.

I want feedback from Go module experts on my proposal above.

PR #391, opened 2019-01-19
fix the EOF issue trying to iterate over pending even when pending is 0

### LINKED ISSUE #390: EOF error returned for a basic Do method use case
## Precondition

Using redis docker image 5.0.3, configured with docker-compose like the following.
```yaml
version: '3'
services:
  redis:
    container_name: "my_redis"
    image: "redis:5.0.3"
    ports:
    - "9100:9100"
```

## Problem

When executing the `redis.Conn.Do` method, consistently getting `EOF` error, even though the issue doesn't occur when I run the command on the redis server directly. At best, `EOF` is not a very meaningful error message in this context.

## Reproducible Scenario

I was facing this in a very basic use case for `SET` command. Below is the minimal code example.

```go
package main

import "github.com/gomodule/redigo/redis"

func main() {
	conn, err := redis.Dial("tcp", "localhost:9100")
	if err != nil {
		panic(err)
	}

	defer conn.Close()

	_, err = conn.Do("SET", "foo", "bar")
	if err != nil {
		panic(err)  // <------ Panics here, the error returned is "EOF"
	}
}
```

sanggonlee:
Fixes #390

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -671,7 +671,7 @@ func (c *conn) DoWithTimeout(readTimeout time.Duration, cmd string, args ...inte
 
 	var err error
 	var reply interface{}
-	for i := 0; i <= pending; i++ {
+	for i := 0; i < pending; i++ {
 		var e error
 		if reply, e = c.readReply(); e != nil {
 			return nil, c.fatal(e)


garyburd:
Return values are not read with this change.

PR #393, opened 2019-01-31
deal with simple strings when scanning string slices

mmasouras:
deal with cases where we get a slice of values, some of which are simple strings and some are byte arrays.

https://groups.google.com/forum/#!topic/redisearch/2-OR4coLaR8


garyburd:
See complete fix with tests in #394.

Issue #403, opened 2019-02-28
Unable to determine wait stats for pools


stevenh:
When Wait is enabled and MaxActive > 0 calls to `pool.Get()` will wait for an available connection.

This is powerful feature but without stats about the stats for these waits its impossible to determine if the pool connection config are tuned correctly.

We need the ability query stats to will allow these values to be correction turned as is done by https://golang.org/pkg/database/sql/#DBStats.

PR #404, opened 2019-02-28
Add wait statistics to pools

stevenh:
Add WaitCount and WaitDuration stats to pool stats to enable tracking of the delays introduced by waiting for connections from the pool.

This uses the same properties as database/sql so its familiar to users.

Addresses: #403


garyburd:
The documentation for the exported fields should state that the values may not be accurate.

stevenh:
Thanks @garyburd updated.

PR #408, opened 2019-03-20
Add Pool Do and DoScript helpers

stevenh:
Add Do and DoScript helpers to Pool which allow individual Do commands to be easily executed while holding a connection for a minimum amount of time.

This is important for apps which mix redis commands with other work, which would otherwise need to implement individual functions or closures which perform the Get / Close methods in order to avoid holding redis connections while not actually using them.

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -253,6 +253,24 @@ func (p *Pool) Close() error {
 	return nil
 }
 
+// Do sends a command to the server and returns the received reply on a connection from the pool.
+// This can be used to easily minimize the time a connection is held from the pool for an individual Conn.Do.
+func (p *Pool) Do(command string, args ...interface{}) (reply interface{}, err error) {
+	c := p.Get()
+	defer c.Close()
+
+	return c.Do(command, args...)
+}
+
+// DoScript evaluates the script on a connection from the pool.
+// This can be used to easily minimize the time a connection is held from the pool for an individual Script.Do.
+func (p *Pool) DoScript(script *Script, args ...interface{}) (interface{}, error) {
+	c := p.Get()
+	defer c.Close()
+
+	return script.Do(c, args...)
+}
+
 func (p *Pool) lazyInit() {
 	// Fast path.
 	if atomic.LoadUint32(&p.chInitialized) == 1 {
--- a/redis/pool_test.go
+++ b/redis/pool_test.go
@@ -445,6 +445,51 @@ func TestPoolMonitorCleanup(t *testing.T) {
 	d.check("", p, 1, 0, 0)
 }
 
+func TestPoolDo(t *testing.T) {
+	d := poolDialer{t: t}
+	p := &redis.Pool{
+		MaxIdle:   2,
+		MaxActive: 2,
+		Dial:      d.dial,
+	}
+	defer p.Close()
+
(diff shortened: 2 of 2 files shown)


garyburd:
An application can easily provide its own convenience functions for executing a single command and single script on a pooled connection. 

These are two of many possible convenience methods that can be added to Pool.   Where's the line between what should be added on Pool or left to the application?  The current line is that Pool does not include convenience methods.

stevenh:
I would tend to agree however the need for these two seems significantly greater than most others I could think of given the impact it can have on the performance of the redis server itself.

This was highlighted to us by the amount of additional connections used by our apps due to the use of the code flows like the following:
```go
func myHelper() error {
   c2 := pool.Get()
   defer c2.Close()
   _, err := c2.Do("SET", "mykey2", "val")
   return err
}

func myFunc() error {
   c1 := pool.Get()
   defer c1.Close()
   if _, err := c1.Do("INC", "mykey"); err != nil {
      return err
   }
   // do some stuff ....
   if err := myFunc2(); err != nil {
       return err
   }
   // do some more stuff ...
   return nil
}
```

In the above we have two problems:
* c1 is held far longer than needed
* When `myFunc2()` is called from `myFunc()` we require two redis connections when one would do.

To your point about the app implementing it, I totally agree we could have a methods in each app to achieve the above e.g.
```go
func PoolDo(p *redis.Pool, cmd string, args ... interface{}) (interface{}, error) {
  c := p.Get()
  defer c.Close()
  return c.Do(cmd, args...)
}
```
This works, its not may lines but it does still have drawbacks:
1. Every app needs to implement it.
2. It's not quite as clean as we now need to `v, err := PoolDo(pool, "INC", "mykey")` vs. `pool.Do("INC", "mykey")`, which while not much more for an app which uses a lot does all add up.
3. Not everyone will identify the need for the above until it's actually causing an issue.

I believe _3._ is the big issue and having the helpers available will avoid this by encouraging _best practice_ and avoid the gotcha of holding connections open then not needed due to the use of `c := pool.Get()` `defer c.Close()` pattern.

Does that make sense?

stevenh:
I forgot to mention we noticed this as it started causing redis server overload, which is why we started investigating and came up with this as a fix.

We could have made this a little a helper, but to my point this wouldn't help other users avoid the same issue in the future.

garyburd:
Don't use defer if it's not right for your scenario.

```
func myFunc() error {
   c1 := pool.Get()
   err := c1.Do("INC", "mykey")
   c1.Close()
   if err != nil {
      return err
   }
   // do some stuff ....
   if err := myFunc2(); err != nil {
       return err
   }
   // do some more stuff ...
   return nil
}

```

stevenh:
That works for the simple case however in the general case `defer` is our preference for a few reasons:
* Manual `Close()` calls can get complex with multiple statements and lead to a missing `Close()` and hence connection leak.
* When part of external processing e.g. via a web request you want to ensure an unexpected panic can't leak.

In essence `defer` is the idiomatic way unless in the hot path and the user really wants to optimise for performance, at which point they accept the risks associated with doing so.

Given this I believe that adding these methods to the standard library will encourage better use patterns that minimise connection scope and all the benefits that brings; thoughts?

Issue #366, opened 2018-09-25
Support Go Modules


garyburd:
This repo was moved from here from garyburd/redigo. At the time of the move, I made some minor breaking changes to the API and tagged a v2.0.0 release.

The v2.0.0 release turned out to be a problem when Go Modules were introduced. It's now clear that I should have deleted previous v1.x.x releases from the repo and tagged a new v1.0.0 release.

I was my intent that this repo have one major semantic version until the next breaking API change is made. Let's call the major semantic version as of today as version N. 

~~It's possible that users who employ a dependency management tool are using the tags copied over from garyburd/redigo. These users are on major version N-1.~~  Edit: These versions do not compile because they have a package import comment for garyburd/redigo.  It's safe to assume that there are no users of these versions.

Users of Go 1.5+  who do not employ a dependency management tool should continue to get the latest of major version N.  This works today and should continue to be supported until Go Modules is widely adopted.

Users who opt into Go Modules should not need to change their code.

Because I expect Go Modules to be widely adopted before major version N+1 of this package is created, it's acceptable to require a dependency management tool to get version N+1.

Adding a go.mod file to a v2.x.x release changes the path of the package to github.com/gomodule/redigo/v2.  This requires a change to all importers of the packages and does not work for users on Go 1.8 and earlier.

See previous discussion in #365 and #363.

ostcar:
I am not sure if I do something wrong or if the current state is broken.

I try to use the new go modules on an existing project. When I call `go build` go installs all dependencies including redigo but fails with

```
../../.gopath/pkg/mod/github.com/garyburd/redigo@v2.0.0+incompatible/redis/pool.go:28:2: use of internal package github.com/gomodule/redigo/internal not allowed
```

Do you know how to solve this or do you have a workaround?

elithrar:
The error refers to garyburd/redigo and not gomodule/redigo (the correct
path). How are you importing the library in your packages?

ostcar:
sry. I found the reason. My question was off topic here so I deleted it.

huaguoping:
Hi, i can not build. get the error:

> /workpath/go/pkg/mod/github.com/garyburd/redigo@v2.0.0+incompatible/redis/pool.go:28:2: use of internal package github.com/gomodule/redigo/internal not allowed

And i got 2 redigo in my go.mod...
> require (
	...
	github.com/garyburd/redigo v2.0.0+incompatible
	github.com/gomodule/redigo v2.0.0+incompatible // indirect
[...]

PR #413, opened 2019-03-21
add mod file

arnottcr:
- [go modules](https://github.com/golang/go/wiki/Modules) will be enabled in 1.13.
- after merge it would be nice if you would tag this repo: e.g. `v1.7.1`
- since there exists an old `v2.0.0` tag, a major subdirectory, `v2`, has been broken out
  - all exported symbols are mapped using type alases with the aide of [`goforward`](https://golang.org/cl/137076)
  - references point to `v1.7.0`
  - probably best to keep the minor.patch in sync, thus I suggest tagging `v2.7.0`
  - this change could be cherry picked as a major branch if preferred
  - if there are concerns about compatibility with other non-module dependency management solutions, you could use `v3` instead and release `v3.7.0`


elithrar:
Please see the discussion in https://github.com/gomodule/redigo/issues/366

We can’t “drop” the v2.0.0 tag as others are likely using it with other dependency management tools (prior to Go Modules).

PR #426, opened 2019-06-19
Fix a bug in pool initialization

yfei1:
`select` clause will do a random pick when multiple cases are ready. This will, for example, make sending a transaction with a canceled context possible to succeed when `p.ch` is ready. This commit changes this behavior to let `get` function always returns a deterministic error when input is a canceled context.


propertyscott:
This breaks the scenario where the context is canceled during the call to get.  Use the following code to deterministically return an error when  `ctx` is canceled before the call to `get`.

    if ctx == nil {
        <-p.ch
    } else if err := ctx.Err(); err != nil {
        return nil, err
    } else {
        select {
        case <-p.ch:
        case <-ctx.Done():
        return nil, ctx.Err()
        }
    }

yfei1:
> if ctx == nil {
>     <-p.ch
> } else if err := ctx.Err(); err != nil {
>     return nil, err
> } else {
>     select {
>     case <-p.ch:
>     case <-ctx.Done():
>     return nil, ctx.Err()
>     }
> }

Thanks for point it out. Done

stevenh:
@yfei1 any update on this?

stevenh:
@yfei1 just checking in on this one, could you please update?

stevenh:
Hi @yfei1 this has been pending fixes for quite some time now, could you please update this otherwise I'm going to have to close this one.

stevenh:
Closing as stale, please reopen if you want to continue work on this.

PR #434, opened 2019-08-29
Add Peek() method to see if there is a message waiting to be read

rwbarber2:
I want to be able to receive all pubsub messages and process them in a single 'transaction' so I need to be able to peek to see if there may be another message incoming such that Receive would not block.

### DIFF
--- a/redis/redis.go
+++ b/redis/redis.go
@@ -43,6 +43,9 @@ type Conn interface {
 
 	// Receive receives a single reply from the Redis server
 	Receive() (reply interface{}, err error)
+
+	// Peek returns true if these is a byte in the Recieve buffer (indicating a message awaits)
+	Peek() bool
 }
 
 // Argument is the interface implemented by an object which wants to control how
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -611,6 +611,10 @@ func (c *conn) Flush() error {
 	return nil
 }
 
+func (c *conn) Peek() bool {
+	return c.br.Buffered() > 0
+}
+
 func (c *conn) Receive() (interface{}, error) {
 	return c.ReceiveWithTimeout(c.readTimeout)
 }
--- a/redis/log.go
+++ b/redis/log.go
@@ -133,6 +133,12 @@ func (c *loggingConn) Send(commandName string, args ...interface{}) error {
 	return err
 }
 
+func (c *loggingConn) Peek() bool {
+	reply := c.Conn.Peek()
+	c.print("Peek", "", nil, reply, nil)
+	return reply
+}
+
 func (c *loggingConn) Receive() (interface{}, error) {
 	reply, err := c.Conn.Receive()
 	c.print("Receive", "", nil, reply, err)
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -538,6 +538,14 @@ func (ac *activeConn) Flush() error {
 	return pc.c.Flush()
 }
 
+func (ac *activeConn) Peek() bool {
+	pc := ac.pc
+	if pc == nil {
+		return false
+	}
+	return pc.c.Peek()
+}
+
 func (ac *activeConn) Receive() (reply interface{}, err error) {
 	pc := ac.pc
 	if pc == nil {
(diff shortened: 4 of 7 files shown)


stevenh:
Is there a reason why receive with timeout doesn't allow you to achieve the same thing?

rwbarber2:
> Is there a reason why receive with timeout doesn't allow you to achieve the same thing?

That was my first approach.  However, ReceiveWithTimeout closes the connection upon timeout which makes its use very inefficient.  My first PR was going to be to detect a timeout error and not close the connection in fatal() but I was concerned about breaking backward compatibility so I changed to using Peek().  Peek() is more efficient than a modified ReceiveWithTimeout solution.

stevenh:
> > Is there a reason why receive with timeout doesn't allow you to achieve the same thing?
> 
> That was my first approach. However, ReceiveWithTimeout closes the connection upon timeout which makes its use very inefficient.

That's interesting, sounds like an issue in itself, as that should only be the case if data has been received i.e. we can't tell the state of the connection.

> My first PR was going to be to detect a timeout error and not close the connection in fatal() but I was concerned about breaking backward compatibility so I changed to using Peek(). Peek() is more efficient than a modified ReceiveWithTimeout solution.

My concern with Peek is that it adds additional requirements to the Conn interface which has a very narrow use case. In addition it's not predictable, as it depends on how the messages are sent and received at the low level, so there's no guarantee how the messages are grouped which sounds like you're relying on?

rwbarber2:
> My concern with Peek is that it adds additional requirements to the Conn interface which has a very narrow use case. In addition it's not predictable, as it depends on how the messages are sent and received at the low level, so there's no guarantee how the messages are grouped which sounds like you're relying on?

Understood.  My current solution limits the modifications to another package (to limit the amount of refactoring).  It Receive(s) all available messages and places them in an array, then processes each one making some decisions, then finally executes the decisions.  Worst case is I refactor the other package to extract messages from a go channel :/

stevenh:
Would you be able to provide some context to why you're looking process them as a group?

rwbarber2:
> Would you be able to provide some context to why you're looking process them as a group?

Sure, each message represents a change to a data structure and making the data structure changes is quick/fast.  But after the changes are made an analysis process must happen that is CPU intensive and takes some time.  So I want to collect all the changes and apply them quickly then start the analysis process once for the batch of changes.  Make sense?

stevenh:
Do the changes always come in groups? 
Does system know which changes are grouped?

rwbarber2:
> Do the changes always come in groups?

The changes do not necessarily come in groups.  Usually they do not come in groups, but when they are grouped it's a 'large' group (dozens).

> Does system know which changes are grouped?

It has no idea when/if there will be a large group or not.

redis/redis.go
```diff
@@ -43,6 +43,9 @@ type Conn interface {
 
 	// Receive receives a single reply from the Redis server
 	Receive() (reply interface{}, err error)
+
+	// Peek returns true if these is a byte in the Recieve buffer (indicating a message awaits)
+	Peek() bool
```
elithrar: As discussed, this would break the API for all current consumers (outside of a version change/Go Modules).

If you were to take this route (not endorsing it either way), creating a `ConnPeeker` interface that embeds `Conn` would be the preferred way to go.

stevenh:
Sounds like the best option would be to have a secondary process step, something like:
```go
t := time.NewTimer(combineTimeout)
var changes bool
for {
     if !t.Stop() {
	<-t.C
     }
     t.Reset(combineTimeout)
     select {
     case <-changes:
          // Apply change
         changes = true
     case <-t.C:
          if changes {
              // No more pending changes 
              // Analyse
              changes = false
         }
     }
}
```

The above guarantees to combine all changes with combineTimeout of each other calling a single analyse, unlike `Peek`, which is what you're looking to achieve?

redis/redis.go
rwbarber2: > As discussed, this would break the API for all current consumers (outside of a version change/Go Modules).
> 
> If you were to take this route (not endorsing it either way), creating a `ConnPeeker` interface that embeds `Conn` would be the preferred way to go.

Good call, I agree with you.  I think I will abandon this direction and go with a secondary internal process.

rwbarber2:
> Sounds like the best option would be to have a secondary process step, something like:
> 
> ```go
> t := time.NewTimer(combineTimeout)
> var changes bool
> for {
>      if !t.Stop() {
> 	<-t.C
>      }
>      t.Reset(combineTimeout)
>      select {
>      case <-changes:
>           // Apply change
>          changes = true
>      case <-t.C:
>           if changes {
>               // No more pending changes 
>               // Analyse
>               changes = false
>          }
>      }
> }
> ```
> 
> The above guarantees to combine all changes with combineTimeout of each other calling a single analyse, unlike `Peek`, which is what you're looking to achieve?

Thank you.  I'm abandoning this method and will move to a secondary process.  Thanks for reviewing this :)

szmcdull:
I really need this Peek() method. I use a pubsub to receive value changes. But I only need that last value. The changes are very rapid but slow to process. So I need a handy way to skip intermediate values but the last one.

PR #440, opened 2019-10-11
gomod

### LINKED ISSUE #366: Support Go Modules

dcormier:
Fixes #366 (I think).

Once merged, [additional tags need to be added](https://github.com/golang/go/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository) to the repo:
- `redis/v3.0.0`
- `redisx/v3.0.0`

This will make this repo support go modules, and follows the [official guidance](https://github.com/golang/go/wiki/Modules#incrementing-the-major-version-when-first-adopting-modules-with-v2-packages) of bumping the major version when there are released tags that are not actually for `go mod`.

It maybe worth [considering not having the `redis` and `redisx` modules in the same repo](https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository). Before taking this change might be a good opportunity to do that since consumers will have to change the import path, anyway.

At the very least, it may be worthwhile to move the `redisx` package out of this repo, since it's intended to be experimental and should probably remain `v0`.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -2,7 +2,7 @@ Redigo
 ======
 
 [![Build Status](https://travis-ci.org/gomodule/redigo.svg?branch=master)](https://travis-ci.org/gomodule/redigo)
-[![GoDoc](https://godoc.org/github.com/gomodule/redigo/redis?status.svg)](https://pkg.go.dev/github.com/gomodule/redigo/redis/v3)
+[![GoDoc](https://godoc.org/github.com/gomodule/redigo/redis?status.svg)](https://godoc.org/github.com/gomodule/redigo/redis)
 
 Redigo is a [Go](http://golang.org/) client for the [Redis](http://redis.io/) database.
 
@@ -28,14 +28,9 @@ Installation
 
 Install Redigo using the "go get" command:
 
-    go get github.com/gomodule/redigo/redis/v3
+    go get github.com/gomodule/redigo/redis
 
 The Go distribution is Redigo's only dependency.
-
-Updating
---------
-
-If you you were using `redigo` prior to the `v3` tag and are using `go mod` you'll need to update your import paths to add the `/v3` suffix as detailed by [Installation](#installation). A community tool [github.com/marwan-at-work/mod](https://github.com/marwan-at-work/mod) helps automate this procedure. See the [repository](https://github.com/marwan-at-work/mod) or the [community tooling FAQ](https://github.com/golang/go/wiki/Modules#what-community-tooling-exists-for-working-with-modules) below for an overview.
 
 Related Projects
 ----------------
--- a/README.markdown
+++ b/README.markdown
@@ -2,7 +2,7 @@ Redigo
 ======
 
 [![Build Status](https://travis-ci.org/gomodule/redigo.svg?branch=master)](https://travis-ci.org/gomodule/redigo)
-[![GoDoc](https://godoc.org/github.com/gomodule/redigo/redis?status.svg)](https://godoc.org/github.com/gomodule/redigo/redis)
+[![GoDoc](https://godoc.org/github.com/gomodule/redigo/redis?status.svg)](https://pkg.go.dev/github.com/gomodule/redigo/redis/v3)
 
 Redigo is a [Go](http://golang.org/) client for the [Redis](http://redis.io/) database.
 
@@ -28,9 +28,14 @@ Installation
 
 Install Redigo using the "go get" command:
 
-    go get github.com/gomodule/redigo/redis
+    go get github.com/gomodule/redigo/redis/v3
 
 The Go distribution is Redigo's only dependency.
+
+Updating
+--------
+
+If you you were using `redigo` prior to the `v3` tag and are using `go mod` you'll need to update your import paths to add the `/v3` suffix as detailed by [Installation](#installation). A community tool [github.com/marwan-at-work/mod](https://github.com/marwan-at-work/mod) helps automate this procedure. See the [repository](https://github.com/marwan-at-work/mod) or the [community tooling FAQ](https://github.com/golang/go/wiki/Modules#what-community-tooling-exists-for-working-with-modules) below for an overview.
 
 Related Projects
 ----------------
--- a/redis/conn_test.go
+++ b/redis/conn_test.go
@@ -28,7 +28,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/redis/v3"
 )
 
 type testConn struct {
--- a/redis/conn_test.go
+++ b/redis/conn_test.go
@@ -28,7 +28,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 )
 
 type testConn struct {
--- a/redis/go.mod
+++ b/redis/go.mod
@@ -1,5 +1,5 @@
-module github.com/gomodule/redigo/redis/v3
+module github.com/gomodule/redigo/redis
 
-go 1.13
+go 1.14
 
 require github.com/stretchr/testify v1.4.0
--- a/redis/go.mod
+++ b/redis/go.mod
@@ -1,0 +1,5 @@
+module github.com/gomodule/redigo/redis/v3
+
+go 1.13
+
+require github.com/stretchr/testify v1.4.0
--- a/redis/go.sum
+++ b/redis/go.sum
@@ -1,0 +1,11 @@
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/testify v1.4.0 h1:2E4SXV/wtOkTonXsotYi4li6zVWxYlZuYNCXe9XRJyk=
+github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81PSLYec5m4=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v2 v2.2.2 h1:ZCJp+EgiOT7lHqUV2J862kp8Qj64Jo6az82+3Td9dZw=
+gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
--- a/redis/pool_test.go
+++ b/redis/pool_test.go
@@ -23,7 +23,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 )
 
 const (
--- a/redis/pool_test.go
+++ b/redis/pool_test.go
@@ -23,7 +23,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/redis/v3"
 )
 
 const (
--- a/redis/pubsub_example_test.go
+++ b/redis/pubsub_example_test.go
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/redis/v3"
 )
 
 // listenPubSubChannels listens for messages on Redis pubsub channels. The
--- a/redis/pubsub_example_test.go
+++ b/redis/pubsub_example_test.go
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 )
 
 // listenPubSubChannels listens for messages on Redis pubsub channels. The
--- a/redis/pubsub_test.go
+++ b/redis/pubsub_test.go
@@ -19,7 +19,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 )
 
 func expectPushed(t *testing.T, c redis.PubSubConn, message string, expected interface{}) {
--- a/redis/pubsub_test.go
+++ b/redis/pubsub_test.go
@@ -19,7 +19,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/redis/v3"
 )
 
 func expectPushed(t *testing.T, c redis.PubSubConn, message string, expected interface{}) {
--- a/redis/redis_test.go
+++ b/redis/redis_test.go
@@ -18,7 +18,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/redis/v3"
 )
 
 type timeoutTestConn int
--- a/redis/redis_test.go
+++ b/redis/redis_test.go
@@ -18,7 +18,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 )
 
 type timeoutTestConn int
--- a/redis/reply_test.go
+++ b/redis/reply_test.go
@@ -21,7 +21,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 )
 
 type valueError struct {
--- a/redis/reply_test.go
+++ b/redis/reply_test.go
@@ -21,7 +21,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/redis/v3"
 )
 
 type valueError struct {
--- a/redis/scan_test.go
+++ b/redis/scan_test.go
@@ -22,7 +22,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 	"github.com/stretchr/testify/require"
 )
 
--- a/redis/scan_test.go
+++ b/redis/scan_test.go
@@ -22,7 +22,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/redis/v3"
 	"github.com/stretchr/testify/require"
 )
 
--- a/redis/script_test.go
+++ b/redis/script_test.go
@@ -20,7 +20,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
+	"github.com/gomodule/redigo/redis/v3"
 )
 
 var (
--- a/redis/script_test.go
+++ b/redis/script_test.go
@@ -20,7 +20,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 )
 
 var (
--- a/redis/zpop_example_test.go
+++ b/redis/zpop_example_test.go
@@ -17,7 +17,7 @@ package redis_test
 import (
 	"fmt"
 
-	"github.com/gomodule/redigo/redis/v3"
+	"github.com/gomodule/redigo/redis"
 )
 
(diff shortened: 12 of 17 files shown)


stevenh:
I may be missing something but why not just add the v3 tag?

dcormier:
The CI failures here are from go1.7 and 1.8. Versions prior to go1.9.7 do not support [minimal module compatibility](https://github.com/golang/go/wiki/Modules#how-are-v2-modules-treated-in-a-build-if-modules-support-is-not-enabled-how-does-minimal-module-compatibility-work-in-197-1103-and-111).

Additionally, at this point, these versions are well outside the window of support from the golang team anyway. [Only the 2 most recent major versions are supported.](https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance)

Is it a requirement to continue to support these older versions?

Songmu:
I'm glad to support Go Modules.

Wouldn't it be okay to place only one go.mod at the top level of the repository like the following. I think that is easier to understand.

https://github.com/gomodule/redigo/commit/654073aa203a96a879aea61427a98c9a30c9935f

In this way, we just only add v3.0.0 tag.

Certainly, redisx is an experimental package, so it's hard to follow the semantic version and keep compatibility. But, isn't this package exceptionally acceptable for incompatible fixes?

dcormier:
> Wouldn't it be okay to place only one go.mod at the top level of the repository like the following. I think that is easier to understand.
>
> 654073a
>
> In this way, we just only add v3.0.0 tag.

It looks like [that would work](https://github.com/golang/go/wiki/Modules#gomod) just fine. Please go that route if it's your preference.

> But, isn't this package exceptionally acceptable for incompatible fixes?

I'm not certain what you mean, here, but I think you're asking if it's OK to not semantically version `redisx` and continue to have incompatible fixes there. If it's in this current repo, I don't see a way to avoid it being semantically versioned because this repo already has version tags on it. If it's moved out to another repo (say, `github.com/gomodule/redisx`), you can make it a go module there without ever tagging it with a version. That way it would be implied to be `v0`, and incompatible fixes would be permitted there. [Russ Cox's guidance on that:](https://github.com/golang/go/issues/24301#issuecomment-371228664)

>v0 is omitted from import paths because - according to [semver](https://semver.org/) - there are no compatibility guarantees at all for those versions. Requiring an explicit v0 element would do little to ensure compatibility; you'd have to say v0.1.2 to be completely precise, updating all import paths on every update of the library. That seems like overkill. Instead we hope that developers will simply look at the list of modules they depend on and be appropriately wary of any v0.x.y versions they find.

stevenh:
Ok I think we've let this one sit long enough. I've just put a PR to remove 1.7.x and 1.8.x validation which had their last release at the beginning of 2017.

If anyone has any objections to this please say now, otherwise I intend to merge: https://github.com/gomodule/redigo/pull/465 and then this PR which I'm hoping @dcormier can rebase.

Please use 👍 or 👎 to express your preference!

dcormier:
Rebased.

dcormier:
Is there anything else preventing this from moving forward?

stevenh:
Just giving people time to respond, its only been a few days; end of the week or beginning of next.

stevenh:
So I've been testing this morning how this impacts existing projects and I must be missing something as without changing consumer import to includes the `/v3` suffix it simply doesn't work.

```
go get -u
go: github.com/stevenh/gomodtest2 upgrade => v3.0.0+incompatible
go: downloading github.com/stevenh/gomodtest2 v3.0.0+incompatible
go: finding module for package github.com/stevenh/gomodtest2/mypkg
go: finding module for package github.com/stevenh/gomodtest2/mypkg
main.go:4:2: module github.com/stevenh/gomodtest2@latest found (v3.0.0+incompatible), but does not contain package github.com/stevenh/gomodtest2/mypkg
```

This is a terrible user experience as it will literally break everyones projects, did I miss something?

dcormier:
No, you didn't miss something. Unfortunately, that is how gomod works. The docs linked in my comment at the beginning of this PR lay it out. Unfortunately, they are lengthy. But since this repo has version tags, the repo has to play by gomod's rules if consumers are to get the version of this package that one would expect.

LasTshaMAN:
> As a reminder, the state this repo is currently in is that if you're using a supported version of go (currently go1.13+) and go a go get -u github.com/gomodule/redigo/redis to try to add this package to your project, you'll be given v2.0.0, from March 15, 2018. Rather than the most recent tag, v1.7.0, from January 19, 2019.

I can corroborate that, tried to `go get` the latest version and ended up with `v2.0.0` :(

stevenh:
I was afraid of that, I think the guide could do with clarifying when it comes to the user side as it gives the impression it will just work which is not the case.

I kind of expected it given the internal URL's had to change but it's definitely not 100% clear.

So given this I think we need to get an update to this to detail the changes people need front an center in the README, seems like the only tool out there that that does this currently is https://github.com/marwan-at-work/mod, there's [thread here about making a core tool](https://github.com/golang/go/issues/32014).

Is this something you have some time do to @dcormier?

dcormier:
I've updated the readme to give the correct import path and docs link for once the `redis/v3.0.0` tag is added to the repo and the new version is [added to pkg.go.dev](https://go.dev/about#adding-a-package).

If you'd like some additional text, feel free to provide some. You can either reply with it (and where you want it) and I'll update the readme in this PR, open another PR against my branch and I'll merge it, or just update it after this PR is merged.

For what it's worth, as a golang module user, I don't expect every module author to be providing instructions on how to make the tools consume the latest major version upgrade. I just expect repos to be tagged properly, and, if it's present, to show the correct import path in their readme or package-level godoc. It's on me, a golang module user, to know that I have to update my import paths to take major version upgrades.

Relevant docs [here](https://github.com/golang/go/wiki/Modules#semantic-import-versioning):

> If the module is version v2 or higher, the major version of the module _must_ be included as a `/vN` at the end of the module paths used in `go.mod` files (e.g., `module github.com/my/mod/v2`, `require github.com/my/mod/v2 v2.0.1`) and in the package import path (e.g., `import "github.com/my/mod/v2/mypkg"`). This includes the paths used in `go get` commands (e.g., `go get github.com/my/mod/v2@v2.0.1`. Note there is both a `/v2` and a `@v2.0.1` in that example. One way to think about it is that the module name now includes the `/v2`, so include `/v2` whenever you are using the module name).

And, in the same section:

> In general, packages with different import paths are different packages. [...] This is also true if different import paths are due to different major versions appearing in the import path. Thus `example.com/my/mod/mypkg` is a different package than `example.com/my/mod/v2/mypkg`, and both may be imported in a single build, which among other benefits helps with diamond dependency problems and also allows a v1 module to be implemented in terms of its v2 replacement or vice versa.

Available versions for a module are discoverable in a repo itself, and on pkg.go.dev ([example](https://pkg.go.dev/github.com/gomodule/redigo/redis?tab=versions), [another](https://pkg.go.dev/github.com/go-playground/validator/v10?tab=versions)).

README.markdown
```diff
@@ -28,7 +28,7 @@ Installation
 
 Install Redigo using the "go get" command:
 
-    go get github.com/gomodule/redigo/redis
+    go get -u github.com/gomodule/redigo/redis/v3
```
stevenh, line 31:
```suggestion
    go get github.com/gomodule/redigo/redis/v3
```
I believe the recommended way is not to use `-u` with `go mod`

README.markdown
```diff
@@ -28,7 +28,7 @@ Installation
...
 
 The Go distribution is Redigo's only dependency.
```
stevenh, line 34: Given we're making a change which will start erroring when people try to update their dependencies I think we should add a upgrading section, to help people on the their way. I would suggest something like the following which is based off the FAQ entry:

Updating
----------
If you you were using `redigo` prior to the `v3` tag and are using `go mod` you'll need to update your import URL's to add the `/v3` suffix as detailed by [Installation](#installation). A community tool [github.com/marwan-at-work/mod](https://github.com/marwan-at-work/mod) helps automate this procedure. See the [repository](https://github.com/marwan-at-work/mod) or the [community tooling FAQ](https://github.com/golang/go/wiki/Modules#what-community-tooling-exists-for-working-with-modules) below for an overview.

Thoughts?

README.markdown
dcormier: [Yup](https://github.com/golang/go/wiki/Modules#updating-pre-existing-install-instructions), OK.

README.markdown
dcormier: [I added it](https://github.com/dcormier/redigo/blob/dc/gomod/README.markdown#updating). I changed "import URL's" to "import paths", but it's otherwise untouched.

stevenh:
Thanks @dcormier I've asked some colleagues to review, make sure I've not missed something.

Once that's done I'll look to get it merged.

stevenh:
Thanks to bmills in the other thread, looks like go 1.14 has an enhancement that's relevant here:
https://golang.org/doc/go1.14#incompatible-versions

With 1.14 and later we could enable `go mod` support without moving to `v3` thoughts?

dcormier:
If it works, that's the easy way, for sure.

stevenh:
Feels like we should go that route, its less compatible needs people to be on 1.14 but with that it totally eliminates the need for switching to v3 which feels like as pain for a huge amount users.

Thoughts?

dcormier:
If it works, it would be cleaner. Let me update this MR, changing it back to a v1 and adding `go.mod`. That would then need to be merged and tagged as at least just a patch release. Then we can see what the behavior is and if we need to do anything else (hopefully not).

dcormier:
OK, that's done. Tagging that as `v1.7.1` should do the trick.

bcmills:
Having the `go.mod` files in subdirectories means that they are effectively separate modules from the original `github.com/gomodule/redigo`, are independently versioned, and require separate version tags (like `redisx/v1.0.0` and `redis/v1.0.0` respectively).

Given that `redisx` and `redis` have essentially the same set of dependencies anyway, I would urge you to reconsider that decision. (Maintaining two modules in the same repo is more than twice the work of maintaining one, and appears to provide little to no benefit to offset that cost here.)

bcmills:
I would also suggest that this should be a minor-version bump, not a patch-version.

Per https://semver.org/#spec-item-6:
> Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.

“Not using modules” may be undesired behavior, but it is arguably not “incorrect”. Perhaps more importantly, adding a `go.mod` file substantially changes downstream users' dependency versions in a way that they are unlikely to expect.

It's ultimately your call, but I would recommend publishing a revert of this as `v1.7.2` and making another run at it as `v1.8.0`.

arnottcr:
While I second what @bcmills said, the `v1.7.1` tag is also not working. It looks like you merged code that required submodule tags, `redisx/v1.0.0`/`redis/v1.0.0`, but then you released a root level tag `v1.7.1`. This removed all the code and is a breaking change: (because the `go.mod` files exclude their subtrees root module)
```sh
[user@localhost ~]$ go get github.com/gomodule/redigo@v1.7.1
[user@localhost ~]$ ls -R ~/go/pkg/mod/github.com/gomodule/redigo@v1.7.1/
/home/user/go/pkg/mod/github.com/gomodule/redigo@v1.7.1/:
LICENSE  README.markdown
```
As expected, `redis` and `redisx` are now `v0.0.0` pseudo-versions:
```sh
[user@localhost ~]$ go get github.com/gomodule/redigo/redis
go: downloading github.com/gomodule/redigo/redis v0.0.0-20200429221454-e14091dffc1b
go: github.com/gomodule/redigo/redis upgrade => v0.0.0-20200429221454-e14091dffc1b
```
As I agree that this repo should probably have only the one root module, can we revert this change and add one `go.mod` at the root, or if you feel strongly cut `redisx/v1.7.1` and `redis/v1.7.1`, since today things are broken.

stevenh:
Thanks guys will get on that now

stevenh:
Ok so I've done the following:
1. reverted the original
2. tagged v1.7.2
3. created go.mod at the route
4. tagged v1.8.0

This still doesn't work
```
go clean -modcache
user@localhost:~/code/test2$ go get github.com/gomodule/redigo@v1.8.0
go: downloading github.com/gomodule/redigo v1.8.0
user@localhost:~/code/test2$ go get github.com/gomodule/redigo/redis
go: downloading github.com/gomodule/redigo/redis v0.0.0-20200429221454-e14091dffc1b
go: github.com/gomodule/redigo/redis upgrade => v0.0.0-20200429221454-e14091dffc1b
```

If I set GOPRIVATE thinks work as expected
```
go clean -modcache
GOPRIVATE=github.com/gomodule/redigo; export GOPRIVATE
user@localhost:~/code/test2$ go clean -modcache
user@localhost:~/code/test2$ GOPRIVATE=github.com/gomodule/redigo; export GOPRIVATE
user@localhost:~/code/test2$ go get github.com/gomodule/redigo@v1.8.0
go: downloading github.com/gomodule/redigo v1.8.0
user@localhost:~/code/test2$ go get github.com/gomodule/redigo/redis
go: found github.com/gomodule/redigo/redis in github.com/gomodule/redigo v1.8.0
```
Seems like something in the GOPROXY system is out of date, any advice?

arnottcr:
So, it looks like golang/go#34189 is a deeper hole than expected. What is happening is that the fact that any version, even a prerelease, of `github.com/gomodule/redigo/redis` exists means that it is preferred by the toolchain over `github.com/gomodule/redigo`.

This seems potentially reasonable for full valid releases, but for alpha, pre-release, and pseudo-versions this feels like a bug, since the mere existence of a commit within the repo, not on master or even tagged, that contains a problematic submodule `go.mod` will taint that subtree forever.

GOPRIVATE works because it goes directly to VCS (read: github); unfortunately, there is no way to unpublish a module, see golang/go#24031, so I am unclear if you have any option other than moving to submodules. That being said, I would advocate for the root module if it is possible; @bcmills, can you speak to alternative resolutions?

stevenh:
More information:
```
go get github.com/gomodule/redigo/redis@v1.8.0
go: found github.com/gomodule/redigo/redis in github.com/gomodule/redigo v1.8.0
```

```
go get github.com/gomodule/redigo/redis@latest
go: downloading github.com/gomodule/redigo/redis v0.0.0-20200429221454-e14091dffc1b
go: github.com/gomodule/redigo/redis latest => v0.0.0-20200429221454-e14091dffc1b
```
So why does the explicit version work but latest doesn't?

Note that the hash latest is picking up (e14091dffc1b) is older than v1.8.0 (898fbd0a5b)

I have the horrible feeling that now the proxy has seen a go.mod at the lower level there's no going back :(

arnottcr:
> So why does the explicit version work but latest doesn't?

It works because you are asking for a package at a given vcs tag. So, at `v1.8.0` there exists the requested package. You are circumventing the _normal_ resolution flow.
 
> I have the horrible feeling that now the proxy has seen a go.mod at the lower level there's no going back :(

Yeah, I was wanting confirmation from @bcmills to see if this was a known issue. But that is the way it looks, once you go submodule you cannot go back? That being said, this exact behaviour does seem like a bug, since an untagged, pre-release `v0.0.0` is preferred over the valid `v1.8.0`. It just leads to a host of edge cases where a branch/commit/pr/pre-release tag exists and it takes precedent over the valid release.

stevenh:
Indeed, any small slip and its impossible to recover, which is far from ideal.

bcmills:
The fix for the accidental nested modules is to remove the packages from the “latest” version of them, so that the `go` command will notice they they don't actually provide the imports it's looking for and keep trying other paths.

One easy way to do that is with a (presumably off-master) commit that provides trivial `go.mod` files in those subdirectories but deletes all of the directories and files. (The proxy will be able to fetch that commit as long as it is reachable from any branch or tag —not just a SemVer tag — and you can fetch it using `go get -d` with the commit hash as the version.)

bcmills:
You might also be able to achieve the same effect by changing the `module` line in the `go.mod` files so that it no longer matches the module path being fetched. (Such a module can only be used in a `replace` directive for the declared path.)

stevenh:
Have tried this approach but didn't see to work.
To clarify I created a branch with just a `redis/go.mod` `redisx/go.mod` no other files this is here:
github.com/gomodule/redigo/commit/393804bd3349bc80396be3ef10afe3ce339e646f

I then used `go get -d` as below
```
user@localhost:~/code/test2$ go get -d github.com/gomodule/redigo/redis@393804bd3349bc80396be3ef10afe3ce339e646f
go: downloading github.com/gomodule/redigo/redis v0.0.0-20200430125810-393804bd3349
go: downloading github.com/gomodule/redigo v1.8.1-0.20200430125810-393804bd3349
go: github.com/gomodule/redigo/redis 393804bd3349bc80396be3ef10afe3ce339e646f => v0.0.0-20200430125810-393804bd3349
user@localhost:~/code/test2$ go get -d github.com/gomodule/redigo/redisx@393804bd3349bc80396be3ef10afe3ce339e646f
go: downloading github.com/gomodule/redigo/redisx v0.0.0-20200430125810-393804bd3349
go: github.com/gomodule/redigo/redisx 393804bd3349bc80396be3ef10afe3ce339e646f => v0.0.0-20200430125810-393804bd3349
```

Then tried to get the mainline (v1.8.0)
```
user@localhost:~/code/test2$ go clean -modcache
user@localhost:~/code/test2$ go get github.com/gomodule/redigo@v1.8.0
go: downloading github.com/gomodule/redigo v1.8.0
user@localhost:~/code/test2$ go get github.com/gomodule/redigo/redis
go: downloading github.com/gomodule/redigo/redis v0.0.0-20200429221454-e14091dffc1b
go: github.com/gomodule/redigo/redis upgrade => v0.0.0-20200429221454-e14091dffc1b
```
This is still picking up the original nested `go.mod` version.

Did this agree with the process you had in mind @bcmills ?

bcmills:
I think there is a problem with the `proxy.golang.org` cache failing to update `@latest` when the module no longer exists at head. I'm going to escalate this to the `proxy.golang.org` maintainers.

bcmills:
Aha! I think I've got it. You can tag commit `393804bd3349bc80396be3ef10afe3ce339e646f` as `redis/v0.0.0-do-not-use` and `redisx/v0.0.0-do-not-use` and then fetch those tags through the proxy.

Then that tagged version will show up in https://proxy.golang.org/github.com/gomodule/redigo/redis/@v/list, and the `go` command will prefer it over https://proxy.golang.org/github.com/gomodule/redigo/redis/@latest.

dcormier:
I apologize for this mess. I know it wasn't working right before (with gomod picking up the old v2.0.0 tag), but I shouldn't have put this forward. I should have just left it as an issue.

Feel free to completely ignore this and not respond to it at all:

&lt;rant&gt;
To be honest, and excuse my language, but gomod is a shitshow. It's just seems _so_ overcomplicated. I mean, look how long [this page](https://github.com/golang/go/wiki/Modules) is!

We're suddenly forced to adopt it in some cases to get packages to work right, but it has so many [gatchas](https://en.wikipedia.org/wiki/Gotcha_(programming)) that it can be tricky to adopt for a package that existed prior to gomod. And if you don't know all the rules, it's easy to break them and end up in a bad situation with no obvious way out. And gomod suddenly changes the behavior of some preexisting mechanisms (like git tags).

Golang users shouldn't be reliant on members of the golang team to jump in and give direction. It's great that they do that, and I very much appreciate it, but that it's needed seems like a bad sign.

Not to mention that it provides no way to discover that a new major upgrade is available for a package.

I miss `dep`.
&lt;/rant&gt;

dcormier:
@bcmills, does `v0.0.0-do-not-use` have a special meaning that's documented somewhere?

bcmills:
@dcormier, to be fair, the Wiki [recommends against](https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository) multi-module repos, and https://blog.golang.org/migrating-to-go-modules explicitly says “If you have existing version tags, you should increment the minor version.”

(And yeah, we know the reference documentation needs work. That's golang/go#33637, and @jayconrod in particular is actively working on it.)

bcmills:
`v0.0.0-do-not-use` is just a [pre-release version](https://semver.org/#spec-item-9) that no one is likely to mistake for a release intended for public consumption. (There is nothing special about it beyond the fact that it is a valid semantic version.)

stevenh:
> Aha! I think I've got it. You can tag commit `393804bd3349bc80396be3ef10afe3ce339e646f` as `redis/v0.0.0-do-not-use` and `redisx/v0.0.0-do-not-use` and then fetch those tags through the proxy.
> 
> Then that tagged version will show up in https://proxy.golang.org/github.com/gomodule/redigo/redis/@v/list, and the `go` command will prefer it over https://proxy.golang.org/github.com/gomodule/redigo/redis/@latest.

To confirm my understanding I need to tag create the two tags `redis/v0.0.0-do-not-use` and `redisx/v0.0.0-do-not-use` temporarily while we perform `go get` to force the proxy to remove the old records and these tags and that commit can then be removed?

bcmills:
Correct. The proxy won't _remove_ the old records per se, but it will permanently cache the existence of those versions (and the contents at those versions), and the `go` command will try them instead of the (also still cached) bogus pseudo-version.

stevenh:
Looks like we have a winner!
```
user@localhost:~/code/test2$ go clean -modcache
user@localhost:~/code/test2$ go get -d github.com/gomodule/redigo
go: downloading github.com/gomodule/redigo v1.8.0
go: github.com/gomodule/redigo upgrade => v1.8.0
```

Thank you so much for your help @bcmills it would have taken ages to fix that without your help!

Feels like this is a new `go mod` FAQ entry :)

domino14:
Help what is happening. 
```
$ go clean -modcache
$ go get -d github.com/gomodule/redigo
go: finding github.com/gomodule/redigo v2.0.0+incompatible
go: downloading github.com/gomodule/redigo v2.0.0+incompatible
go: extracting github.com/gomodule/redigo v2.0.0+incompatible
``` That is the wrong version right?

**Comment by bcmills** [comment]
@domino14, you need to be using Go 1.14 or higher for `go get` to ignore the `+incompatible` version.

**Comment by stevenh** [comment]
@domino14 your path is wrong you want: 
```
go get github.com/gomodule/redigo/redigo
```

Issue #439, opened 2019-10-07
Pass context to TestOnBorrow


dcormier:
I like to add tracing to my code. This depends on being able to pass a context around. To that end, I wish `TestOnBorrow` was passed the context that's passed to `GetContext`. Perhaps this would require a new function so it's not a breaking change: `TestOnBorrowContext`, or something similar.

stevenh:
Indeed this would require a new method.

dcormier:
Uh, so, was a new method added? The PR was closed without merging, resulting in this issue being closed.

stevenh:
Yes you can see the comments on the https://github.com/gomodule/redigo/pull/442

stevenh:
@dcormier just flagging this was reopened in case you didn't see the notification.

Also mentioning that I'm thinking of a v2 which would help clean up the existing interfaces, so would be interested in peoples feedback on use cases like this we should consider when designing that.

PR #442, opened 2019-11-01
[WIP] Create function like TestOnBorrow that includes context

MorrisLaw:
This is an initial PR to address issue #439

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -141,6 +141,10 @@ type Pool struct {
 	// closed.
 	TestOnBorrow func(c Conn, t time.Time) error
 
+	// TestOnBorrowWithContext is the same as TestOnBorrow, but includes
+	// the context.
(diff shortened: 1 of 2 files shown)


MorrisLaw:
> the ctx with a timeout, which would be the desired use, will expire and resulting in all future calls failing.

I agree @stevenh , seems kind of pointless then... Should we just close this PR out?

stevenh:
Closing based on timeout and other feedback.

dcormier:
> I'm not sure this can ever be of use if I'm honest as this is setup on the pool so the ctx with a timeout, which would be the desired use, will expire and resulting in all future calls failing.

But wouldn't it be called with the `context.Context` passed to `GetContext`? I haven't dug into the code, but I would assume that's when `TestOnBorrow` would be called, so the context used should be correct. And if the context is canceled or timed out, then it should still be correct.

stevenh:
I take it back, @dcormier is correct.

Issue #444, opened 2019-11-05
RedisArg: use it in conjunction with `AddFlat`


cyx:
## Context

Given this scenario:

```
type User struct {
    CreatedAt Time `redis:"created_at"`
}

type Time struct {
    time.Time
}

func (t Time) RedisArg() interface{} {
  return t.Unix()
}
```

We'd expect that doing `redis.Args{}.AddFlat(&User{Time: Time{time.Now()}})` would utilize the underlying `RedisArg` -- but currently it does the default marshaling strategy for `interface{}` instead.

It does seem like the diff is relatively small in doing this, e.g.:

```diff
-                       args = append(args, fs.name, fv.Interface())
+                       if arg, ok := fv.Interface().(Argument); ok {
+                               args = append(args, fs.name, arg.RedisArg())
+                       } else {
+                               args = append(args, fs.name, fv.Interface())
+                       }
```

Does this seem like a good change?

stevenh:
This looks sensible to me, a PR which does this including test would be most appreciated :)

cyx:
@stevenh please see the PR over at #446

PR #446, opened 2019-11-08
addflat: respect RedisArg presence

cyx:
## Context

Given this scenario:

```
type User struct {
    CreatedAt Time `redis:"created_at"`
}

type Time struct {
    time.Time
}

func (t Time) RedisArg() interface{} {
  return t.Unix()
}
```

We'd have the custom `Time.RedisArg` be called when doing `AddFlat`.

Closes https://github.com/gomodule/redigo/issues/444


stevenh:
Thanks for the PR @cyx

PR #453, opened 2019-12-26
add func Uint64Map Uint64s

xuanskyer:
add func Uint64Map Uint64s

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -479,8 +479,8 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
 	return positions, nil
 }
 
-// Same as Int64s, Uint64s is a helper that converts an array command reply to a []uint64.
-// If err is not equal to nil, then Uint64s returns nil, err. Nil array
+// Uint64s is a helper that converts an array command reply to a []uint64.
+//If err is not equal to nil, then Uint64s returns nil, err. Nil array
 // items are stay nil. Uint64s returns an error if an array item is not a
 // bulk string or nil.
 func Uint64s(reply interface{}, err error) ([]uint64, error) {
@@ -502,10 +502,10 @@ func Uint64s(reply interface{}, err error) ([]uint64, error) {
 }
 
 
-// Same as Int64Map , Uint64Map is a helper that return a  map[uint64]uint64 data. 
+// Uint64Map is a helper that return a  map[string]uint64 data.
 //The HGETALL commands return replies in this format.
 // Requires an even number of values in result.
-func Uint64Map(result interface{}, err error) (map[uint64]uint64, error) {
+func Uint64Map(result interface{}, err error) (map[string]uint64, error) {
 	values, err := Values(result, err)
 	if err != nil {
 		return nil, err
@@ -513,17 +513,17 @@ func Uint64Map(result interface{}, err error) (map[uint64]uint64, error) {
 	if len(values)%2 != 0 {
 		return nil, errors.New("redigo: Uint64Map expects even number of values result")
 	}
-	m := make(map[uint64]uint64, len(values)/2)
+	m := make(map[string]uint64, len(values)/2)
 	for i := 0; i < len(values); i += 2 {
-		key, err := Uint64(values[i], nil)
-		if err != nil {
-			return nil, err
+		key, ok := values[i].([]byte)
+		if !ok {
+			return nil, errors.New("redigo: Uint64Map key not a bulk string value")
 		}
 		value, err := Uint64(values[i+1], nil)
 		if err != nil {
 			return nil, err
 		}
-		m[key] = value
+		m[string(key)] = value
 	}
 	return m, nil
 }
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -502,8 +502,8 @@ func Uint64s(reply interface{}, err error) ([]uint64, error) {
 }
 
 
-// Uint64Map is a helper that converts an array of strings (alternating key, value) into a map[string]uint64.
-//The HGETALL commands return replies in this format.
+// Uint64Map is a helper that converts an array of strings (alternating key, value)
+// into a map[string]uint64. The HGETALL commands return replies in this format.
 // Requires an even number of values in result.
 func Uint64Map(result interface{}, err error) (map[string]uint64, error) {
 	values, err := Values(result, err)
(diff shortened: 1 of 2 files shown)


redis/reply.go
```diff
@@ -477,3 +477,44 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
...
+		return nil, err
+	}
+	if len(values)%2 != 0 {
+		return nil, errors.New("redigo: Uint64Map expects even number of values result")
+	}
+	m := make(map[uint64]uint64, len(values)/2)
+	for i := 0; i < len(values); i += 2 {
+		key, err := Uint64(values[i], nil)
+		if err != nil {
+			return nil, errors.New("redigo: Uint64Map key not a bulk uint4 value")
```
stevenh:
```suggestion
			return nil, err
```
I know that other areas of the code use this style, but I think its better to not loose the original error which includes more detail, thoughts?

If not I believe the `bulk uint4` should actually be `bulk string`

redis/reply.go
```diff
@@ -477,3 +477,44 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
 	}
 	return positions, nil
 }
+
+func Uint64s(reply interface{}, err error) ([]uint64, error) {
```
stevenh: Small nit, all the other `XXXs(...)` methods are together, same with the `XXXMap(...)` so be nice to keep each of these in the same area too.

stevenh: Missing doc comment.

redis/reply.go
```diff
@@ -477,3 +477,44 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
...
+			result[i] = n
+			return err
+		default:
+			return fmt.Errorf("redigo: unexpected element type for Uint64s, got type %T", v)
+		}
+	})
+	return result, err
+}
+
+func Uint64Map(result interface{}, err error) (map[uint64]uint64, error) {
```
stevenh: Missing doc comment.

stevenh:
Hi @xuanskyer could you address the comments above and rebase to fix the conflicts please?

xuanskyer:
> Hi @xuanskyer could you address the comments above and rebase to fix the conflicts please?

done

redis/reply.go
```diff
@@ -479,6 +479,55 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
 	return positions, nil
 }
 
+// Same as Int64s, Uint64s is a helper that converts an array command reply to a []uint64.
```
stevenh, line 482:
```suggestion
// Uint64s is a helper that converts an array command reply to a []uint64.
```

redis/reply.go
```diff
@@ -479,6 +479,55 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
...
+			return err
+		default:
+			return fmt.Errorf("redigo: unexpected element type for Uint64s, got type %T", v)
+		}
+	})
+	return result, err
+}
+
+
+// Same as Int64Map , Uint64Map is a helper that return a  map[uint64]uint64 data.
```
stevenh, line 505: I think this one will lead to confusion, I would expect `map[string]uint64` not `map[uint64]uint64` as `Uint64Map` does.

This feels like more of a special case than we should include in this library if this was the real intention?

As per above no need to refer to the `Same as...` and this isn't the format `HGETALL` returns as that's strings, sure it could work but for the most part not, so this could lead to confusion.

Also missing a test for this method.

redis/reply_test.go
```diff
@@ -68,6 +68,16 @@ var replyTests = []struct {
 		ve(redis.Int64s([]interface{}{int64(4), int64(5)}, nil)),
 		ve([]int64{4, 5}, nil),
 	},
+	{
+		"uint64s([uint64, uint64])",
+		ve(redis.Uint64s([]interface{}{uint64(4), uint64(5)}, nil)),
```
stevenh, line 73: Can we include max values here: `math.MaxUint64`

redis/reply.go
xuanskyer: In projects, sometimes I really want to return map[uint64][uint64]. 
maybe I should change the func name `Uint64Map` ?
Or do you have any other suggestions？

stevenh: You may yes but I would say its much less common to have field name that's an integer, I can't say we have any usages for that here.

xuanskyer: > You may yes but I would say its much less common to have field name that's an integer, I can't say we have any usages for that here.
> 
> So I would change this to map[string]uint64

ok, got it.

stevenh:
I think you forgot to push the update there @xuanskyer

redis/reply.go
```diff
@@ -479,6 +479,55 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
 	return positions, nil
 }
 
+// Uint64s is a helper that converts an array command reply to a []uint64.
+//If err is not equal to nil, then Uint64s returns nil, err. Nil array
```
stevenh, line 483:
```suggestion
// If err is not equal to nil, then Uint64s returns nil, err. Nil array
```

redis/reply.go
```diff
@@ -479,6 +479,55 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
...
+			return fmt.Errorf("redigo: unexpected element type for Uint64s, got type %T", v)
+		}
+	})
+	return result, err
+}
+
+
+// Uint64Map is a helper that return a  map[string]uint64 data.
+//The HGETALL commands return replies in this format.
+// Requires an even number of values in result.
```
stevenh, line 507:
```suggestion
// Uint64Map is a helper that converts an array of strings (alternating key, value)
// into a map[string]uint64. The HGETALL commands return replies in this format.
// Requires an even number of values in result.
```

redis/reply.go
```diff
@@ -479,6 +479,55 @@ func Positions(result interface{}, err error) ([]*[2]float64, error) {
...
+			return fmt.Errorf("redigo: unexpected element type for Uint64s, got type %T", v)
+		}
+	})
+	return result, err
+}
+
+
+// Uint64Map is a helper that converts an array of strings (alternating key, value) into a map[string]uint64.
+//The HGETALL commands return replies in this format.
+// Requires an even number of values in result.
```
stevenh, line 507:
```suggestion
// Uint64Map is a helper that converts an array of strings (alternating key, value)
// into a map[string]uint64. The HGETALL commands return replies in this format.
// Requires an even number of values in result.
```

PR #457, opened 2020-02-11
Simplify string reply check

cuonglm:
The go compiler is smart enough to optimize string([]byte) conversion.

```
name       old time/op    new time/op    delta
DoPing-12    29.2µs ± 3%    29.3µs ± 4%   ~     (p=0.730 n=9+9)

name       old alloc/op   new alloc/op   delta
DoPing-12     0.00B          0.00B        ~     (all equal)

name       old allocs/op  new allocs/op  delta
DoPing-12      0.00           0.00        ~     (all equal)
```

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -541,11 +541,11 @@ func (c *conn) readReply() (interface{}, error) {
 	}
 	switch line[0] {
 	case '+':
-		switch {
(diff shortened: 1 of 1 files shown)


stevenh:
Thanks for the PR

PR #470, opened 2020-04-26
Handle `ctx` cancellation promptly (when waiting for a vacant connection from the pool)

LasTshaMAN:
I've also stumbled on [this one](https://github.com/gomodule/redigo/pull/426) on a project I've been working on recently. It might lead to pretty nasty scenarios when service under load uses `redigo`.

For some context you can check out [this reddit tread](https://www.reddit.com/r/golang/comments/4ibs68/why_is_the_select_statement_nondeterministic/).

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -361,7 +361,7 @@ func (p *Pool) lazyInit() {
 
 // waitVacantConn waits for a vacant connection in pool if waiting
 // is enabled and pool size is limited, otherwise returns instantly.
-// If `ctx` expires before that, an error is returned.
+// If ctx expires before that, an error is returned.
 //
 // If there were no vacant connection in the pool right away it returns the time spent waiting
 // for that connection to appear in the pool.
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -359,7 +359,8 @@ func (p *Pool) lazyInit() {
 	p.mu.Unlock()
 }
 
-// waitVacantConn waits until at least one vacant connection gets spotted in the pool.
+// waitVacantConn waits for a vacant connection in pool if waiting
+// is enabled and pool size is limited, otherwise returns instantly.
 // If `ctx` expires before that, an error is returned.
 //
 // If there were no vacant connection in the pool right away it returns the time spent waiting
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -199,27 +199,10 @@ func (p *Pool) Get() Conn {
 // If the function completes without error, then the application must close the
 // returned connection.
 func (p *Pool) GetContext(ctx context.Context) (Conn, error) {
-	// Handle limit for p.Wait == true.
-	var waited time.Duration
-	if p.Wait && p.MaxActive > 0 {
-		p.lazyInit()
-
-		// wait indicates if we believe it will block so its not 100% accurate
-		// however for stats it should be good enough.
-		wait := len(p.ch) == 0
-		var start time.Time
-		if wait {
-			start = time.Now()
-		}
-		select {
-		case <-p.ch:
-		case <-ctx.Done():
-			err := ctx.Err()
-			return errorConn{err}, err
-		}
-		if wait {
-			waited = time.Since(start)
-		}
+	// Wait until there is a vacant connection in the pool.
+	waited, err := p.waitVacantConn(ctx)
+	if err != nil {
+		return nil, err
 	}
 
 	p.mu.Lock()
@@ -374,6 +357,52 @@ func (p *Pool) lazyInit() {
 		atomic.StoreUint32(&p.chInitialized, 1)
 	}
 	p.mu.Unlock()
+}
+
+// waitVacantConn waits until at least one vacant connection gets spotted in the pool.
+// If `ctx` expires before that, an error is returned.
+//
+// If there were no vacant connection in the pool right away it returns the time spent waiting
+// for that connection to appear in the pool.
+func (p *Pool) waitVacantConn(ctx context.Context) (waited time.Duration, err error) {
+	if !p.Wait {
+		return 0, nil
+	}
+	if p.MaxActive <= 0 {
+		return 0, nil
+	}
+
+	p.lazyInit()
+
+	// wait indicates if we believe it will block so its not 100% accurate
+	// however for stats it should be good enough.
+	wait := len(p.ch) == 0
+	var start time.Time
+	if wait {
+		start = time.Now()
+	}
+
+	if ctx == nil {
+		<-p.ch
+	} else {
+		select {
+		case <-p.ch:
+			// Additionally check that context hasn't expired while we were waiting,
+			// because `select` picks a random `case` if several of them are "ready".
+			select {
+			case <-ctx.Done():
+				return 0, ctx.Err()
+			default:
+			}
+		case <-ctx.Done():
+			return 0, ctx.Err()
+		}
+	}
+
+	if wait {
+		return time.Since(start), nil
+	}
+	return 0, nil
 }
 
 func (p *Pool) dial(ctx context.Context) (Conn, error) {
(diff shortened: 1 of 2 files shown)


LasTshaMAN:
I also should note the following:

_if you are using `redis.Pool.Dial` and not `redis.Pool.DialContext` (that handles `ctx` cancellation  properly) your `go-routine` might still get stuck when dialing for a connection (even if the context you are using when calling `redis.Pool.GetContext` is cancelled). If your `redis.Pool.Dial` doesn't implement timeouts your `go-routine` will stuck forever. So it is something to think about._ 

Any reasons on why `redigo` doesn't provide default `ctx`-based `redis.Pool.DialContext` implementation similar to https://github.com/gomodule/redigo/blob/master/redis/conn.go#L179-L251 ? If there is no inherent problems with it, I could drop a PR ...

redis/pool.go
```diff
@@ -299,34 +299,61 @@ func (p *Pool) lazyInit() {
...
+		select {
+		case <-p.ch:
+			// Additionally check that context hasn't expired while we were waiting,
+			// because `select` picks a random `case` if several of them are "ready".
 			select {
-			case <-p.ch:
 			case <-ctx.Done():
-				return nil, ctx.Err()
+				return 0, ctx.Err()
 			}
```
stevenh, line 337: This will actually block until `ctx.Done()` as you're missing the `default` case from the `select`

redis/pool.go
```diff
@@ -299,34 +299,61 @@ func (p *Pool) lazyInit() {
...
+// - (as its first return value) the time it spent waiting, if there were no vacant connection in the pool right away
+// or zero otherwise
+// - (as its second return value) error, if any
+func (p *Pool) waitVacantConn(ctx context.Context) (time.Duration, error) {
+	if !p.Wait {
+		return 0, nil
+	}
+	if p.MaxActive <= 0 {
+		return 0, nil
+	}
```
stevenh, line 315: These could be combined

redis/pool.go
```diff
@@ -299,34 +299,61 @@ func (p *Pool) lazyInit() {
 	p.mu.Unlock()
 }
 
-// get prunes stale connections and returns a connection from the idle list or
-// creates a new connection.
-func (p *Pool) get(ctx context.Context) (*poolConn, error) {
+// waitVacantConn waits while at least one vacant connection appears in the pool.
```
stevenh, line 302: This doesn't reflect what the code actually does.

redis/pool.go
```diff
@@ -299,34 +299,61 @@ func (p *Pool) lazyInit() {
...
+// If `ctx` expires before that, an error is returned.
+//
+// waitVacantConn returns:
+// - (as its first return value) the time it spent waiting, if there were no vacant connection in the pool right away
+// or zero otherwise
+// - (as its second return value) error, if any
```
stevenh, line 308: It's not idiomatic to describe a return like this, see some of the core libraries for examples e.g.
https://golang.org/pkg/io/#Copy

redis/pool.go
```diff
@@ -299,34 +299,61 @@ func (p *Pool) lazyInit() {
...
+		return time.Since(start), nil
+	}
+	return 0, nil
+}
+
+// get prunes stale connections and returns a connection from the idle list or
+// creates a new connection.
+func (p *Pool) get(ctx context.Context) (*poolConn, error) {
+
+	// Wait while there is a vacant connection in the pool.
```
stevenh, line 353: The opposite is actually true

redis/pool.go
```diff
@@ -299,34 +299,61 @@ func (p *Pool) lazyInit() {
...
+	if wait {
+		return time.Since(start), nil
+	}
+	return 0, nil
+}
+
+// get prunes stale connections and returns a connection from the idle list or
+// creates a new connection.
+func (p *Pool) get(ctx context.Context) (*poolConn, error) {
+
```
stevenh, line 352:
```suggestion

```

redis/pool.go
```diff
@@ -299,34 +299,61 @@ func (p *Pool) lazyInit() {
...
+	if ctx == nil {
+		<-p.ch
+	} else {
+		select {
+		case <-p.ch:
+			// Additionally check that context hasn't expired while we were waiting,
+			// because `select` picks a random `case` if several of them are "ready".
 			select {
-			case <-p.ch:
```
stevenh, line 322: While I don't mind the refactor its arguably simpler to just add the additional select in here and be done:
```go
select {
case <-ctx.Done()
    return nil, ctx.Err()
default:
}
```

redis/pool.go
LasTshaMAN: Sorry, it's my poor English I guess ;)

Would `Wait until there is a vacant connection in the pool.` sound about right ?

redis/pool.go
LasTshaMAN: How about `waitVacantConn waits until at least one vacant connection gets spotted in the pool.` ?

redis/pool.go
LasTshaMAN: Actually I've split them "on purpose" for readability reasons, but if you think it reads better when combined I can return it to the original version.

redis/pool.go
LasTshaMAN: yeah, you are right, and your test-suite actually spotted it as well, I just didn't have spare time to check it out.

Fixed.

redis/pool.go
LasTshaMAN: fixed

redis/pool.go
LasTshaMAN: Changed `func` signature and its description a bit.

redis/pool.go
LasTshaMAN: I've changed some bits (as you've suggested), 

if you still prefer to have the original version of the code with "extra" select, I'll remove refactoring from PR

stevenh:
Hi @LasTshaMAN could you rebase is to fix the conflicts. Will re-review once that's done.

LasTshaMAN:
Sure, rebased.

@stevenh please check out this PR once again.

redis/pool.go
```diff
@@ -376,6 +359,52 @@ func (p *Pool) lazyInit() {
...
+//
+// If there were no vacant connection in the pool right away it returns the time spent waiting
+// for that connection to appear in the pool.
+func (p *Pool) waitVacantConn(ctx context.Context) (waited time.Duration, err error) {
+	if !p.Wait {
+		return 0, nil
+	}
+	if p.MaxActive <= 0 {
+		return 0, nil
+	}
```
stevenh, line 373:
```suggestion
	if !p.Wait || p.MaxActive <= 0 {
		// No wait or no connection limit.
		return 0, nil
	}
```

redis/pool.go
```diff
@@ -376,6 +359,52 @@ func (p *Pool) lazyInit() {
 	p.mu.Unlock()
 }
 
+// waitVacantConn waits until at least one vacant connection gets spotted in the pool.
+// If `ctx` expires before that, an error is returned.
```
stevenh, line 363:
```suggestion
// If ctx expires before that, an error is returned.
```
Doc strings shouldn't use markdown syntax.

redis/pool.go
```diff
@@ -376,6 +359,52 @@ func (p *Pool) lazyInit() {
 	p.mu.Unlock()
 }
 
+// waitVacantConn waits until at least one vacant connection gets spotted in the pool.
```
stevenh, line 362:
```suggestion
// waitVacantConn waits for a vacant connection in pool if waiting
// is enabled and pool size is limited, otherwise returns instantly.
```

redis/pool.go
stevenh: yep lets combine and add a comment as per suggestion below.

LasTshaMAN: fixed

LasTshaMAN:
I'll squash commits upon merging as well.

PR #474, opened 2020-05-01
fix: Argument processing on pointers

stevenh:
Fix Argument interface processing on struct pointer types which was broken by #428 and not quite corrected by #446.


stevenh:
@cyx and @wych42 could you review / test this change in your setup as it could well impacts you both.

wych42:
It looks good to me, and does not break my current setup&usage.

PR #476, opened 2020-05-10
Add DialContext function

### LINKED ISSUE #475: Add DialContext function
I want Context to be used when dialing.
So I want DialContext.

By default you should use Dialer.DialContext.
https://golang.org/pkg/net/#Dialer.DialContext

178inaba:
Fixes #475.

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -78,7 +78,6 @@ type dialOptions struct {
 	readTimeout  time.Duration
 	writeTimeout time.Duration
 	dialer       *net.Dialer
-	dial         func(network, addr string) (net.Conn, error)
 	dialContext  func(ctx context.Context, network, address string) (net.Conn, error)
 	db           int
 	password     string
@@ -86,19 +85,6 @@ type dialOptions struct {
 	useTLS       bool
 	skipVerify   bool
 	tlsConfig    *tls.Config
-}
-
-func newDialOptions(options []DialOption) dialOptions {
-	do := dialOptions{
-		dialer: &net.Dialer{
-			KeepAlive: time.Minute * 5,
-		},
-	}
-	for _, option := range options {
-		option.f(&do)
-	}
-
-	return do
 }
 
 // DialReadTimeout specifies the timeout for reading a single command reply.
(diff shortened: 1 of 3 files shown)


redis/conn.go
```diff
@@ -127,6 +142,15 @@ func DialNetDial(dial func(network, addr string) (net.Conn, error)) DialOption {
 	}}
 }
 
+// DialNetDialContext specifies a custom dial function with context for creating TCP
+// connections, otherwise a net.Dialer customized via the other options is used.
+// DialNetDialContext overrides DialConnectTimeout and DialKeepAlive.
+func DialNetDialContext(dialContext func(ctx context.Context, network, addr string) (net.Conn, error)) DialOption {
```
stevenh, line 148: A more idiomatic name for this would be DialWithContext

redis/conn.go
```diff
@@ -78,6 +79,7 @@ type dialOptions struct {
 	writeTimeout time.Duration
 	dialer       *net.Dialer
 	dial         func(network, addr string) (net.Conn, error)
```
stevenh, line 81: As this is internal we could always use dialContext with a context.Background if need.

redis/conn.go
```diff
@@ -127,6 +142,15 @@ func DialNetDial(dial func(network, addr string) (net.Conn, error)) DialOption {
 	}}
 }
 
+// DialNetDialContext specifies a custom dial function with context for creating TCP
+// connections, otherwise a net.Dialer customized via the other options is used.
+// DialNetDialContext overrides DialConnectTimeout and DialKeepAlive.
+func DialNetDialContext(dialContext func(ctx context.Context, network, addr string) (net.Conn, error)) DialOption {
+	return DialOption{func(do *dialOptions) {
+		do.dialContext = dialContext
```
stevenh, line 150: We could simplify things by only using a dialer with context, as mentioned above.

178inaba: Fix!

redis/conn.go
178inaba: Fix!

redis/conn.go
178inaba: Do I have to be consistent with DialNetDial?
https://pkg.go.dev/github.com/gomodule/redigo/redis?tab=doc#DialNetDial

Dialer uses DialContext.
https://golang.org/pkg/net/#Dialer.DialContext

There is an example to use WithContext when it takes the context itself as an argument.
https://golang.org/pkg/net/http/#Request.WithContext

stevenh: You are right that `DialNetDialContext` follows the precedent of `DialNetDial` but I think it's quite a mouth full.

What do you think users would prefer?

178inaba: What about `DialContextFunc`?

178inaba:
@stevenh I added a test!

178inaba:
@stevenh Hello!
Can you review this pull request again?

PR #478, opened 2020-05-10
Fix Pool.Get to simple code

### LINKED ISSUE #477: Pool.Get code can be simple
The pull request #478 makes the code simple.

178inaba:
Fixes #477.

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -185,6 +185,7 @@ func NewPool(newFn func() (Conn, error), maxIdle int) *Pool {
 // getting an underlying connection, then the connection Err, Do, Send, Flush
 // and Receive methods return that error.
 func (p *Pool) Get() Conn {
+	// GetContext returns errorConn in the first argument when an error occurs.
 	c, _ := p.GetContext(context.Background())
 	return c
 }
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -211,15 +211,11 @@ func (p *Pool) GetContext(ctx context.Context) (Conn, error) {
 		if wait {
 			start = time.Now()
 		}
-		if ctx == nil {
-			<-p.ch
-		} else {
-			select {
-			case <-p.ch:
-			case <-ctx.Done():
-				err := ctx.Err()
-				return errorConn{err}, err
-			}
+		select {
+		case <-p.ch:
+		case <-ctx.Done():
+			err := ctx.Err()
+			return errorConn{err}, err
 		}
 		if wait {
 			waited = time.Since(start)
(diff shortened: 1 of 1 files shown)


redis/pool.go
```diff
@@ -185,11 +185,8 @@ func NewPool(newFn func() (Conn, error), maxIdle int) *Pool {
 // getting an underlying connection, then the connection Err, Do, Send, Flush
 // and Receive methods return that error.
 func (p *Pool) Get() Conn {
-	pc, err := p.get(nil)
-	if err != nil {
-		return errorConn{err}
-	}
-	return &activeConn{p: p, pc: pc}
+	c, _ := p.GetContext(context.Background())
```
stevenh, line 188: Your discarding the error here, which isn't safe

redis/pool.go
```diff
@@ -315,14 +312,10 @@ func (p *Pool) get(ctx context.Context) (*poolConn, error) {
...
-			<-p.ch
-		} else {
-			select {
-			case <-p.ch:
-			case <-ctx.Done():
-				return nil, ctx.Err()
-			}
+		select {
+		case <-p.ch:
+		case <-ctx.Done():
```
stevenh, line 317: This will cause a panic if ctx is nil e.g.
https://play.golang.org/p/F4NjaHeF4bi

redis/pool.go
178inaba: Like `Get`, `GetContext` returns `errorConn` in case of an error, so I think it's okay.

https://github.com/gomodule/redigo/blob/129455dc28ae7cb56466318ec259422d14851207/redis/pool.go#L203-L209

redis/pool.go
178inaba: Thank you!
I reverted!

redis/pool.go
stevenh: We should add a comment to clarify that as its not immediately clear.

178inaba: Added a comment!

178inaba:
@stevenh I have integrated Pool.get into Pool.GetContext.
I think you can follow because the following godoc comments are described in GetContext.
> The provided Context must be non-nil. If the context expires before the connection is complete, an error is returned. Any expiration on the context will not affect the returned connection.

stevenh:
> @stevenh I have integrated Pool.get into Pool.GetContext.
> I think you can follow because the following godoc comments are described in GetContext.
> 
> > The provided Context must be non-nil. If the context expires before the connection is complete, an error is returned. Any expiration on the context will not affect the returned connection.

Now you have eliminated `get(...)` then I would agree

PR #480, opened 2020-05-12
Added dialoptions and Url parsing to connect using Redis ACL system

### LINKED ISSUE #479: Add support for ACL new `AUTH username password`

filipecosta90:
- fixes #479 
- follows the described in [AUTH](https://redis.io/commands/auth) to connect to a Redis 6.0 instance, or greater, that is using the Redis ACL system.
- adds godoc example for further reference

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -79,6 +79,7 @@ type dialOptions struct {
 	dialer       *net.Dialer
 	dial         func(network, addr string) (net.Conn, error)
 	db           int
+	username     string
 	password     string
 	clientName   string
 	useTLS       bool
@@ -139,6 +140,14 @@ func DialDatabase(db int) DialOption {
 func DialPassword(password string) DialOption {
 	return DialOption{func(do *dialOptions) {
 		do.password = password
+	}}
+}
+
+// DialUsername specifies the username to use when connecting to
+// the Redis server when Redis ACLs are used.
+func DialUsername(username string) DialOption {
+	return DialOption{func(do *dialOptions) {
+		do.username = username
 	}}
 }
 
@@ -227,7 +236,12 @@ func Dial(network, address string, options ...DialOption) (Conn, error) {
 	}
 
 	if do.password != "" {
-		if _, err := c.Do("AUTH", do.password); err != nil {
+		authArgs := make([]interface{}, 0, 2)
+		if do.username != "" {
+			authArgs = append(authArgs, do.username)
+		}
+		authArgs = append(authArgs, do.password)
+		if _, err := c.Do("AUTH", authArgs...); err != nil {
 			netConn.Close()
 			return nil, err
 		}
@@ -285,7 +299,7 @@ func DialURL(rawurl string, options ...DialOption) (Conn, error) {
 	if u.User != nil {
 		password, isSet := u.User.Password()
 		if isSet {
-			options = append(options, DialPassword(password))
+			options = append(options, DialUsername(u.User.Username()), DialPassword(password))
 		}
 	}
 
--- a/redis/conn_test.go
+++ b/redis/conn_test.go
@@ -38,10 +38,14 @@ type testConn struct {
 	writeDeadline time.Time
 }
 
-func (*testConn) Close() error                         { return nil }
-func (*testConn) LocalAddr() net.Addr                  { return nil }
-func (*testConn) RemoteAddr() net.Addr                 { return nil }
-func (c *testConn) SetDeadline(t time.Time) error      { c.readDeadline = t; c.writeDeadline = t; return nil }
+func (*testConn) Close() error         { return nil }
+func (*testConn) LocalAddr() net.Addr  { return nil }
+func (*testConn) RemoteAddr() net.Addr { return nil }
+func (c *testConn) SetDeadline(t time.Time) error {
+	c.readDeadline = t
+	c.writeDeadline = t
+	return nil
+}
 func (c *testConn) SetReadDeadline(t time.Time) error  { c.readDeadline = t; return nil }
 func (c *testConn) SetWriteDeadline(t time.Time) error { c.writeDeadline = t; return nil }
 
@@ -600,7 +604,9 @@ var dialURLTests = []struct {
 	r           string
 	w           string
 }{
-	{"password", "redis://x:abc123@localhost", "+OK\r\n", "*2\r\n$4\r\nAUTH\r\n$6\r\nabc123\r\n"},
+	{"password", "redis://:abc123@localhost", "+OK\r\n", "*2\r\n$4\r\nAUTH\r\n$6\r\nabc123\r\n"},
+	{"username and password", "redis://user:password@localhost", "+OK\r\n", "*3\r\n$4\r\nAUTH\r\n$4\r\nuser\r\n$8\r\npassword\r\n"},
+	{"username", "redis://x:@localhost", "+OK\r\n", ""},
 	{"database 3", "redis://localhost/3", "+OK\r\n", "*2\r\n$6\r\nSELECT\r\n$1\r\n3\r\n"},
 	{"database 99", "redis://localhost/99", "+OK\r\n", "*2\r\n$6\r\nSELECT\r\n$2\r\n99\r\n"},
 	{"no database", "redis://localhost/", "+OK\r\n", ""},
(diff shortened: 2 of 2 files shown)


redis/conn.go
```diff
@@ -142,6 +144,22 @@ func DialPassword(password string) DialOption {
...
+	return DialOption{func(do *dialOptions) {
+		do.username = username
+	}}
+}
+
+// DialUseACL specifies whether Redis ACL system should be used when connecting to the
+// server.
+func DialUseACL(useACL bool) DialOption {
+	return DialOption{func(do *dialOptions) {
+		do.useACL = useACL
```
stevenh, line 159: This doesn't seem to have any use apart from when combined with username, so is that sufficient i.e. if you the username is set its used?

redis/conn.go
```diff
@@ -227,7 +245,14 @@ func Dial(network, address string, options ...DialOption) (Conn, error) {
 	}
 
 	if do.password != "" {
-		if _, err := c.Do("AUTH", do.password); err != nil {
+		authArgs := []interface{}{}
```
stevenh, line 248:
```suggestion
		authArgs := make([]interface{}, 0, 2)
```

redis/conn.go
```diff
@@ -283,6 +308,7 @@ func DialURL(rawurl string, options ...DialOption) (Conn, error) {
 	address := net.JoinHostPort(host, port)
 
 	if u.User != nil {
+		options = append(options, DialUsername(u.User.Username()))
```
stevenh, line 311: Does this make sense if password isn't set?

redis/conn.go
filipecosta90: @stevenh I believe it might be necessary since we dont want to break compatibility wih previous url parsing ( that contained the user in the URL ). So I would stick to enforcing the user to aknowledge that he wants to use ACLs. WDYT?

redis/conn.go
filipecosta90: You're right. It will have no effect on the request. changed in the new commit

redis/conn.go
stevenh: Given that would be trivial to fix, what I would say is a broken config anyway I think I'd prefer for simple use. In the case of a parsed url, you would need to enable this anyway as there is nothing else to tell otherwise.

filipecosta90: after asking for more advice youre right ( no  point on having this extra flag =) ). removing it

filipecosta90: solved in the latest commit :+1:

filipecosta90: > Given that would be trivial to fix, what I would say is a broken config anyway, I think I'd prefer to go for simple use.
> 
> In the case of a parsed url, you would need to enable this anyway as there is nothing else to tell otherwise.

you're right @stevenh. was making something more complex and buggy than it was required. fixed =)

filipecosta90:
@stevenh last fix based on your comments =)

redis/conn.go
```diff
@@ -444,12 +458,12 @@ func (c *conn) writeArg(arg interface{}, argumentTypeOK bool) (err error) {
 type protocolError string
 
 func (pe protocolError) Error() string {
-	return fmt.Sprintf("redigo: %s (possible server error or unsupported concurrent read by application)", string(pe))
+	return fmt.Sprintf("redigo: %s (possible server error or unsupported concurrent r by application)", string(pe))
```
platinummonkey, line 461: accidental change? here and below

platinummonkey: https://github.com/filipecosta90/redigo/pull/1 to fix this 👍

stevenh, line 461:
```suggestion
	return fmt.Sprintf("redigo: %s (possible server error or unsupported concurrent read by application)", string(pe))
```

redis/conn.go
```diff
@@ -444,12 +458,12 @@ func (c *conn) writeArg(arg interface{}, argumentTypeOK bool) (err error) {
...
 }
 
 // readLine reads a line of input from the RESP stream.
 func (c *conn) readLine() ([]byte, error) {
-	// To avoid allocations, attempt to read the line using ReadSlice. This
+	// To avoid allocations, attempt to r the line using ReadSlice. This
```
stevenh, line 466:
```suggestion
	// To avoid allocations, attempt to read the line using ReadSlice. This
```

redis/conn.go
```diff
@@ -630,7 +644,7 @@ func (c *conn) ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err
 	// unsubscribing from all channels, we do not decrement pending to a
 	// negative value.
 	//
-	// The pending field is decremented after the reply is read to handle the
+	// The pending field is decremented after the reply is r to handle the
```
stevenh, line 647:
```suggestion
	// The pending field is decremented after the reply is read to handle the
```

redis/conn_test.go
```diff
@@ -276,7 +280,7 @@ func TestReadString(t *testing.T) {
 	// n is value of bufio.defaultBufSize
 	const n = 4096
 
-	// Test read string lengths near bufio.Reader buffer boundaries.
+	// Test r string lengths near bufio.Reader buffer boundaries.
```
stevenh, line 283:
```suggestion
	// Test read string lengths near bufio.Reader buffer boundaries.
```

redis/pool.go
```diff
@@ -468,7 +468,7 @@ func (ac *activeConn) Close() error {
 		pc.c.Send("UNSUBSCRIBE")
 		pc.c.Send("PUNSUBSCRIBE")
 		// To detect the end of the message stream, ask the server to echo
-		// a sentinel value and read until we see that value.
+		// a sentinel value and r until we see that value.
```
stevenh, line 471:
```suggestion
		// a sentinel value and read until we see that value.
```

redis/redis.go
```diff
@@ -55,7 +55,7 @@ type Argument interface {
 }
 
 // Scanner is implemented by an object which wants to control its value is
-// interpreted when read from Redis.
+// interpreted when r from Redis.
```
stevenh, line 58:
```suggestion
// interpreted when read from Redis.
```

redis/redis.go
```diff
@@ -66,11 +66,11 @@ type Scanner interface {
 }
 
 // ConnWithTimeout is an optional interface that allows the caller to override
-// a connection's default read timeout. This interface is useful for executing
+// a connection's default r timeout. This interface is useful for executing
```
stevenh, line 69:
```suggestion
// a connection's default read timeout. This interface is useful for executing
```

redis/redis.go
```diff
@@ -66,11 +66,11 @@ type Scanner interface {
...
 // the BLPOP, BRPOP, BRPOPLPUSH, XREAD and other commands that block at the
 // server.
 //
-// A connection's default read timeout is set with the DialReadTimeout dial
+// A connection's default r timeout is set with the DialReadTimeout dial
```
stevenh, line 73:
```suggestion
// A connection's default read timeout is set with the DialReadTimeout dial
```

redis/redis.go
```diff
@@ -83,18 +83,18 @@ type ConnWithTimeout interface {
 	Conn
 
 	// Do sends a command to the server and returns the received reply.
-	// The timeout overrides the read timeout set when dialing the
+	// The timeout overrides the r timeout set when dialing the
```
stevenh, line 86:
```suggestion
	// The timeout overrides the read timeout set when dialing the
```

redis/redis.go
```diff
@@ -83,18 +83,18 @@ type ConnWithTimeout interface {
...
 	// connection.
 	DoWithTimeout(timeout time.Duration, commandName string, args ...interface{}) (reply interface{}, err error)
 
 	// Receive receives a single reply from the Redis server. The timeout
-	// overrides the read timeout set when dialing the connection.
+	// overrides the r timeout set when dialing the connection.
```
stevenh, line 91:
```suggestion
	// overrides the read timeout set when dialing the connection.
```

redis/redis.go
```diff
@@ -83,18 +83,18 @@ type ConnWithTimeout interface {
...
 	ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err error)
 }
 
 var errTimeoutNotSupported = errors.New("redis: connection does not support ConnWithTimeout")
 
-// DoWithTimeout executes a Redis command with the specified read timeout. If
+// DoWithTimeout executes a Redis command with the specified r timeout. If
```
stevenh, line 97:
```suggestion
// DoWithTimeout executes a Redis command with the specified read timeout. If
```

redis/redis.go
```diff
@@ -105,7 +105,7 @@ func DoWithTimeout(c Conn, timeout time.Duration, cmd string, args ...interface{
 	return cwt.DoWithTimeout(timeout, cmd, args...)
 }
 
-// ReceiveWithTimeout receives a reply with the specified read timeout. If the
+// ReceiveWithTimeout receives a reply with the specified r timeout. If the
```
stevenh, line 108:
```suggestion
// ReceiveWithTimeout receives a reply with the specified read timeout. If the
```

redis/conn.go
platinummonkey: bump @filipecosta90

filipecosta90: sorry for the delay @platinummonkey =) on it

filipecosta90: solved in the latest commit

filipecosta90:
@stevenh can you check it again please. Revise all suggested fixes and cleaned the PR =)

platinummonkey:
bump @stevenh 🙏

ullumullu:
Bump - to avoid forking-off if this is merged in the next couple of days. Verified on our end that the changes make sense and are working.

platinummonkey:
Agreed, we forked and it works fine.

stevenh:
Sorry guys been a busy few weeks.

redis/conn_test.go
```diff
@@ -675,6 +681,36 @@ func TestDialTLSSKipVerify(t *testing.T) {
...
+	_, err := redis.Dial("tcp", "localhost:6379",
+		redis.DialUsername("user"),
+		redis.DialPassword("password"),
+		dialTestConn(pingResponse, &buf))
+	if err != nil {
+		t.Fatal("dial error:", err)
+	}
+	if err != nil {
+		t.Fatal("dial error:", err)
+	}
```
nashikb, line 695: Hmm.

filipecosta90: nice catch :) will fix

PR #494, opened 2020-06-20
fix: GetContext must return errorConn in the first argument whenever …

aclisp:
…an error occurs.


stevenh:
Thanks for the PR @aclisp

PR #496, opened 2020-06-22
Add `WaitTimeout` to prevent `pool.Wait` blocking forever.

knadh:
We use `pool.Wait` to serve high throughput requests.  `Wait` blocks forever, and there are scenarios where this behaviour can deadlock. For example `defer c.Close()` in nested functions can race and block forever. To avoid defer, one then has to `Get()` and `Close()` in multiple places, missing `Close()` in subtle ways. These have bitten us hard.

This PR adds an optional `pool.WaitTimeout` duration that returns `ErrPoolExhausted` when `Wait` is true without blocking forever.

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -21,6 +21,7 @@ import (
 	"crypto/sha1"
 	"errors"
 	"io"
+	"math"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -34,9 +35,10 @@ var (
 
 var nowFunc = time.Now // for testing
 
-// ErrPoolExhausted is returned from a pool connection method (Do, Send,
+// ErrPoolExhausted is returned from a pool connection method (Get, Do, Send,
 // Receive, Flush, Err) when the maximum number of database connections in the
-// pool has been reached.
+// pool has been reached or when the pool does not return a connection within
+// the WaitTimeout period when Wait is true.
 var ErrPoolExhausted = errors.New("redigo: connection pool exhausted")
 
 var (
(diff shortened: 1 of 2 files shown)


stevenh:
Looks like this could be achieved by simply using `pool.GetContext(...)`

knadh:
I didn't consider context deadlines because of [this](https://github.com/gomodule/redigo/blob/master/redis/pool.go#L384) condition.

stevenh:
> I didn't consider context deadlines because of [this](https://github.com/gomodule/redigo/blob/master/redis/pool.go#L384) condition.

Not sure how that comes into it, previous ctx could only be nil if you used `pool.Get()` instead of `pool.GetContext(ctx)`.

Since #470 it can never be nil so that if block should be removed.

knadh:
Understood. 

If this PR is amdended add `WaitTimeout` implicitly to `Get()` using `context.WithTimeout`, could that be considered? Would be nice to get away by simple adding `WaitTimeout` during initialisation without having to refactor the rest across a large number of codebases.

stevenh:
The problem with that is it has other implications, so I don't think its a simple as that.

knadh:
Just to confirm, If one doesn't set a `WaitTimeout` duration, there will be no change in `Get()'s` current behaviour.

stevenh:
Actually in your current implementation you're leaking a timer on every request.

If I'm honest this feels like a very niche case created issues internal to your use case, where it would be better to solve the problem there that it would be to do this, which seems more like a workaround.

Happy to discuss more, but seems like ctx is the better way to solve the issue.

If your code base uses `pool.Get()` everywhere there's nothing to stop you creating a slim wrapper to applies your timeout for you.

knadh:
I was indeed proposing updating the PR with `context.WithTimeout()` implicitly inside `Get()`. But yes, I get your point. Rather use `GetContext()` than introduce a new config variable.

PR #497, opened 2020-06-23
Revert "Update README.markdown"

stevenh:
Reverts gomodule/redigo#488

I read the diff backwards, so this should not have been merged as the original was correct.


stevenh:
@ppphp are you OK with this?

ppphp:
> @ppphp are you OK with this?

Sure!

PR #515, opened 2020-09-17
fix: typo in Float64 documentation

hallazzang:


redis/reply.go
```diff
@@ -118,7 +118,7 @@ func Uint64(reply interface{}, err error) (uint64, error) {
 
 // Float64 is a helper that converts a command reply to 64 bit float. If err is
 // not equal to nil, then Float64 returns 0, err. Otherwise, Float64 converts
-// the reply to an int as follows:
+// the reply to a float as follows:
```
stevenh, line 121:
```suggestion
// the reply to a float64 as follows:
```

stevenh (changes requested):
Thanks for the PR @hallazzang I've put a suggestion as the type is actually a `float64` not a `float` so if you could do that and add a full comment to the commit, example below, that would be great!

Correct a cut and past error in the Float64 documentation which listed the return type as int instead of float64

redis/reply.go
hallazzang: @stevenh Thanks, I updated the commit

PR #521, opened 2020-10-02
Implement client side caching 

### LINKED ISSUE #513: About 6.0 feature: client side caching
https://redis.io/topics/client-side-caching
Do we have plan to support this feature?

holykol:
**This PR implements server-assisted client side caching [introduced](https://redis.io/topics/client-side-caching) in Redis 6.0**

This is done entirely at a pool level by running a separate goroutine that subscribes to invalidation broadcasts.
Command responses are, again, cached by middleware at a pool level.

The implementation is very naive and I wanted to make sure I'm going in the right direction.

Not yet implemented:
* Cache size limit
* Broadcasting mode
* Opt-in caching
* Keeping track of TTL
* Caching other types

closes #513

### DIFF
--- a/redisx/cacher.go
+++ b/redisx/cacher.go
@@ -1,0 +1,235 @@
+package redisx
+
+import (
+	"context"
+	"strings"
+	"sync"
+
+	"github.com/gomodule/redigo/redis"
+)
+
+type cachingStrategy int
+
+const (
+	// Tracking strategy lets server to keep track of keys request by client
+	Tracking cachingStrategy = iota
+	// Broadcast strategy allows client to manually subscribe
+	// to invalidation broadcasts for interested keys
+	Broadcast
+)
+
+// Cacher implements client-side caching first appeared in Redis 6.0
+type Cacher struct {
+	Getter   ConnGetter
+	Strategy cachingStrategy
+
+	cache sync.Map // Actual cache. Not yet replaceble
+
+	mu  sync.Mutex // Guard for fields below
+	cid int        // Connection ID of invalidator process
+}
+
+// Matcher interface is used to determine whether key should be cached.
+// Allows for custom caching logic.
+type Matcher interface {
+	Match(key string) bool
+}
+
+type prefixMatcher []string
+
+// NewPrefixMatcher is a helper function that matches all keys with
+// supplied prefixes
+func NewPrefixMatcher(prefixes []string) Matcher {
+	return prefixMatcher(prefixes)
+}
+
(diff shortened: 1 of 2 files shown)


holykol:
Can someone review?

stevenh (commented):
Given not everyone will be using caching and in fact redis 6 for some time, I'm wondering if this should be implemented by a separate sub package?

This would ensure we don't bloat the core pool code with features not everyone will use or want.

What do you think?

holykol:
@stevenh Great idea. I'm a little embarrassed I haven't thought of that before I started trying to fit more code into `redis.Pool`. Give me some time.

holykol:
Didn't mean to close this, but fine. Take a look at now @stevenh

holykol:
OK, I've gave up on trying to come up with abstraction for caching other data types. It's too much work for potentially useless feature:
1. Implementing this feature basically means re-implementing big chunk of Redis functionality locally.
2. Hashes, lists and sets change too fast to benefit from caching.
3. My guess is that most people using Cacher will just store json/protobuf/gob/bare serialized data structures and will not care about other types.

@stevenh Can this be merged with it's current feature set?

stevenh:
Just a quick note to say I haven't had chance to look at this yet, hopefully soon

redisx/cacher.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+// Stats returns aggregated stats about local cache
+func (c *Cacher) Stats() CacheStats {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	size, hits, misses := c.cache.Stats()
+	return CacheStats{size, hits, misses}
+}
```
TwinProduction, line 302:
```suggestion
func (c *Cacher) Stats() CacheStats {
	c.mu.Lock()
	size, hits, misses := c.cache.Stats()
	c.mu.Unlock()
	return CacheStats{size, hits, misses}
}
```

Should probably avoid deferring operations on the mutex in a cache, since deferring adds an extra ~200ns of latency.
Normally, 200ns would be extremely negligible, but for a cache, 200ns matters.

holykol: It's actually like 4ns difference. A quick search reveals that defer performance has been [improved](https://golang.org/doc/go1.14#runtime) in Go 1.14 . I guess there was some overhead before that version.
Also no allocation happens here so everything runs as fast as it can get.

TwinProduction: My bad, didn't know about that.  Yea, 4ns is definitely negligible then.

PR #523, opened 2020-10-12
use IndexByte improve scan bench

imxyb:
bench result:
```
benchmark                  old ns/op     new ns/op     delta
BenchmarkScanStruct-12     1259          1083          -13.98%

benchmark                  old allocs     new allocs     delta
BenchmarkScanStruct-12     20             17             -15.00%

benchmark                  old bytes     new bytes     delta
BenchmarkScanStruct-12     696           600           -13.79%
```


imxyb:
btw, this bench is based on not using structSpecCache.

stevenh (approved):

PR #524, opened 2020-10-13
remove the err == nil because the err is always equal nil

imxyb:


stevenh:
Thanks for the PR @imxyb :)

PR #525, opened 2020-10-13
use zero copy improve string and byte convert

imxyb:
bench result:

```
BenchmarkIntForByte-12     34.3          19.8          -42.27%

benchmark                  old allocs     new allocs     delta
BenchmarkIntForByte-12     1              0              -100.00%

benchmark                  old bytes     new bytes     delta
BenchmarkIntForByte-12     3             0             -100.00%



benchmark              old ns/op     new ns/op     delta
BenchmarkString-12     14.9          2.55          -82.89%

benchmark              old allocs     new allocs     delta
BenchmarkString-12     1              0              -100.00%

benchmark              old bytes     new bytes     delta
BenchmarkString-12     4             0             -100.00%
```

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -17,8 +17,10 @@ package redis
 import (
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
 	"time"
+	"unsafe"
 )
 
 // ErrNil indicates that a reply value is nil.
@@ -45,7 +47,7 @@ func Int(reply interface{}, err error) (int, error) {
 		}
 		return x, nil
 	case []byte:
-		n, err := strconv.ParseInt(string(reply), 10, 0)
+		n, err := strconv.ParseInt(bytesToString(reply), 10, 0)
 		return int(n), err
 	case nil:
 		return 0, ErrNil
@@ -72,7 +74,7 @@ func Int64(reply interface{}, err error) (int64, error) {
 	case int64:
 		return reply, nil
 	case []byte:
-		n, err := strconv.ParseInt(string(reply), 10, 64)
+		n, err := strconv.ParseInt(bytesToString(reply), 10, 64)
 		return n, err
 	case nil:
 		return 0, ErrNil
@@ -106,7 +108,7 @@ func Uint64(reply interface{}, err error) (uint64, error) {
 		}
 		return uint64(reply), nil
 	case []byte:
-		n, err := strconv.ParseUint(string(reply), 10, 64)
+		n, err := strconv.ParseUint(bytesToString(reply), 10, 64)
 		return n, err
 	case nil:
 		return 0, ErrNil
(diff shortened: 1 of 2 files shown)


stevenh (changes requested):
Thanks for the PR, this seems like a risky change based on the comments I found here:
https://github.com/golang/go/issues/25484

This refers to the fact this would be more easy when https://github.com/golang/go/issues/19367 is merged so I think we should wait for that.

stevenh:
Looks like the proposal https://github.com/golang/go/issues/53003 has been accepted recently so we can look to use those when they are implemented.

stevenh:
I believe the compiler in go 1.21 is doing this optimisation by default as there no difference between `string(reply)` and `unsafe.String(unsafe.SliceData(b), len(b))` when running the following benchmark.

```go
func BenchmarkReplyHelpers(b *testing.B) {
	c, err := dial()
	require.NoError(b, err)
	defer c.Close()

	_, err = c.Do("SET", "k1", "1")
	require.NoError(b, err)

	b.Run("String", func(b *testing.B) {
		reply, err := c.Do("GET", "k1")
		require.NoError(b, err)

		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			_, err = redis.String(reply, err)
		}
		b.StopTimer()

		// Outside the loop, so the compiler cannot optimize the function call away.
		require.NoError(b, err)
	})
	b.Run("Int", func(b *testing.B) {
		reply, err := c.Do("GET", "k1")
		require.NoError(b, err)

		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			_, err = redis.String(reply, err)
		}
		b.StopTimer()

		// Outside the loop, so the compiler cannot optimize the function call away.
		require.NoError(b, err)
	})
	b.Run("Int64", func(b *testing.B) {
		reply, err := c.Do("GET", "k1")
		require.NoError(b, err)

		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			_, err = redis.Int64(reply, err)
		}
		b.StopTimer()

		// Outside the loop, so the compiler cannot optimize the function call away.
		require.NoError(b, err)
	})
	b.Run("Uint64", func(b *testing.B) {
		reply, err := c.Do("GET", "k1")
		require.NoError(b, err)

		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			_, err = redis.Uint64(reply, err)
		}
		b.StopTimer()

		// Outside the loop, so the compiler cannot optimize the function call away.
		require.NoError(b, err)
	})
	b.Run("Float64", func(b *testing.B) {
		reply, err := c.Do("GET", "k1")
		require.NoError(b, err)

		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			_, err = redis.Float64(reply, err)
		}
		b.StopTimer()

		// Outside the loop, so the compiler cannot optimize the function call away.
		require.NoError(b, err)
	})
	b.Run("Bool", func(b *testing.B) {
		reply, err := c.Do("GET", "k1")
		require.NoError(b, err)

		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			_, err = redis.Bool(reply, err)
		}
		b.StopTimer()

		// Outside the loop, so the compiler cannot optimize the function call away.
		require.NoError(b, err)
	})
}
```

PR #526, opened 2020-10-14
remove testify dependence

chenjie199234:
remove testify dependence
make this package clean


stevenh (changes requested):
Sorry @chenjie199234 but I prefer the clean tests as created when using testify.

chenjie199234:
@stevenh there are only 2 place useing the testify.why have to use a testing framework.is there any plan to rewrite all testing with testify?

PR #527, opened 2020-10-15
fix: open connection count

### LINKED ISSUE #520: Pool fails to manage number of open connection?
I've got many context deadline exceeded error from Pool.GetContext.

I suspect it fails to manage number of open connection.
It acquires from p.ch here
https://github.com/gomodule/redigo/blob/6f8b8987f003d0b10dd72673c477b054c9189d2c/redis/pool.go#L385
but should we release it back to p.ch here?
https://github.com/gomodule/redigo/blob/6f8b8987f003d0b10dd72673c477b054c9189d2c/redis/pool.go#L390

otherwise available items in p.ch would decrease and eventually couldn't get new one?

rhnvrm:
This commit releases back the connection in the special
case, so that correct count is always maintained.

Closes #520

### DIFF
--- a/redis/pool_test.go
+++ b/redis/pool_test.go
@@ -825,6 +825,32 @@ func TestWaitPoolGetContext(t *testing.T) {
 	defer c.Close()
 }
 
+func TestWaitPoolGetContextIssue520(t *testing.T) {
+	d := poolDialer{t: t}
+	p := &redis.Pool{
+		MaxIdle:   1,
+		MaxActive: 1,
+		Dial:      d.dial,
+		Wait:      true,
+	}
+	defer p.Close()
+	ctx1, cancel1 := context.WithTimeout(context.Background(), 1*time.Nanosecond)
+	defer cancel1()
+	c, err := p.GetContext(ctx1)
+	if err != context.DeadlineExceeded {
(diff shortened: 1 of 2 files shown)


redis/pool_test.go
```diff
@@ -825,6 +825,30 @@ func TestWaitPoolGetContext(t *testing.T) {
...
+func TestWaitPoolGetContextIssue520(t *testing.T) {
+	d := poolDialer{t: t}
+	p := &redis.Pool{
+		MaxIdle:   1,
+		MaxActive: 1,
+		Dial:      d.dial,
+		Wait:      true,
+	}
+	defer p.Close()
+	ctx1, _ := context.WithTimeout(context.Background(), 1*time.Nanosecond)
```
stevenh, line 837: This leaks a cancel see:
https://travis-ci.org/github/gomodule/redigo/builds/736087424

stevenh (changes requested):
Thanks for the first time PR, couple of little issues with the test to fix go vet.

rhnvrm:
@stevenh sorry for the delay, have pushed the context cancellation change to resolve the `go vet` issue in the test.

stevenh:
Thanks @rhnvrm 👍

PR #530, opened 2020-10-28
feat: Add TLS Handshake timeout support

### LINKED ISSUE #509: GetContext never returns in case of TLS setting is ON for Redis Server
I have setup local Redis Server and enabled redis.DialUseTLS(true) with its password using redis.DialPassword by mistake. This server doesn't support TLS. I am creating pool connection and everything was fine but while getting the connection value from from the pool using GetContext() method, it got stuck into infinite loop at Handshake() API.  Ideally, API should return some error or atleast time out. 

All my time out values setup properly like IdleTimeout, MaxIdle, DialConnectTimeout, DialReadTimeout and DialWriteTimeout.

stevenh:
Add support for configuring a timeout for TLS Handshake call via DialTLSHandshakeTimeout DialOption. If no option is specified then the default timeout is 10 seconds.

Also:
* Add a default connect timeout of 30 seconds matching that of net/http.

Fixes #509

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -257,21 +257,18 @@ func DialContext(ctx context.Context, network, address string, options ...DialOp
 		}
 
 		tlsConn := tls.Client(netConn, tlsConfig)
-
-		errc := make(chan error, 2)
-		var timer *time.Timer // for cancelling TLS handshake
+		errc := make(chan error, 2) // buffered so we don't block timeout or Handshake
 		if d := do.tlsHandshakeTimeout; d != 0 {
-			timer = time.AfterFunc(d, func() {
+			timer := time.AfterFunc(d, func() {
 				errc <- tlsHandshakeTimeoutError{}
 			})
+			defer timer.Stop()
 		}
 		go func() {
-			if timer != nil {
-				defer timer.Stop()
-			}
 			errc <- tlsConn.Handshake()
 		}()
 		if err := <-errc; err != nil {
+			// Timeout or Handshake error.
 			netConn.Close() // nolint: errcheck
 			return nil, err
 		}
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -266,11 +266,10 @@ func DialContext(ctx context.Context, network, address string, options ...DialOp
 			})
 		}
 		go func() {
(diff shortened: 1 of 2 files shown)


redis/conn.go
```diff
@@ -238,10 +257,26 @@ func DialContext(ctx context.Context, network, address string, options ...DialOp
 		}
 
 		tlsConn := tls.Client(netConn, tlsConfig)
-		if err := tlsConn.Handshake(); err != nil {
-			netConn.Close()
+
+		errc := make(chan error, 2)
```
mdanzinger, line 261: Why is this channel buffered if you're never actually consuming both events? Since both the timer's function / handshake are being performed in a goroutine, isn't it safe to just use an unbuffered channel?

mdanzinger (approved):
LGTM, small comment on the use of a buffered channel.

redis/conn.go
mdanzinger: On second thought, I am guessing this is to avoid a leak where the goroutine is blocked trying to send to the channel?

redis/conn.go
```diff
@@ -238,10 +257,26 @@ func DialContext(ctx context.Context, network, address string, options ...DialOp
...
+			timer = time.AfterFunc(d, func() {
+				errc <- tlsHandshakeTimeoutError{}
+			})
+		}
+		go func() {
+			err := tlsConn.Handshake()
+			if timer != nil {
+				timer.Stop()
+			}
+			errc <- err
```
danmrichards, line 273: Could be cleaned up a little by deferring the timer stop (should be nearly zero overhead for this since 1.14) and sending the handshake result straight down the channel.
```suggestion
			if timer != nil {
				defer timer.Stop()
			}
			errc <- tlsConn.Handshake()
```

redis/conn.go
stevenh: correct

stevenh:
@danmrichards I have cleaned up the timer cleanup in a slightly different way, could you re-review please.

danmrichards (approved):

razor-1:
great, thanks for adding this! it would be nice to have a new tag published some time soon

Issue #351, opened 2018-06-27
Add DoWithCancel


garyburd:
Add DoWithCancel helper function after Redis releases the [CLIENT UNBLOCK](https://github.com/antirez/redis/commits/client-unblock) command.

```
// DoWithCancel executes cmd with args. If the context is canceled, then
// DoWithCancel unblocks c by executing the CLIENT UNBLOCK command using a
// connection returned from the get function.  Example:
//
//  result, err := redis.Values(redis.DoWithCancel(c, ctx, pool.Get, "BRPOP", "key", 0))
func DoWithCancel(c Conn, ctx context.Context, get func() Conn, cmd string, args ...interface{}) (interface{}, error) {
	id, err := redis.Int(redis.Do("CLIENT", "ID"))
	if err != nil {
		return err
	}
	stop := make(chan struct{})
	wait := make(chan struct{})

	go func() {
		defer close(wait)
		select {
		case <-stop:
		case <-ctx.Done():
			c := get()
			defer c.Close()
			_, err := c.Do("CLIENT", "UNBLOCK", id)
			// TODO: handle error, possibly by closing the network connection on outer c
		}
	}()

	r, err := c.Do(cmd, args...)
	close(stop)
	<-wait
	return r, err
}
```

hnlq715:
+1

stevenh:
One of the interesting challenges with this is knowing if unblock is supported, as allowing this command if not could result in unexpected results. 

This could be handled with the error check and subsequent closing of connections but that's not obvious to the API consumer, as timeouts are usually an non-standard event it will take some identifying that this is the case.

A possible solution could be to have the pool be version aware.

The questions is given it's an edge case do could it cause a real issue for the server admin, and if not do we care?

PR #537, opened 2020-11-20
add DoContext and ReceiveContext,use context to control the life

chenjie199234:
context is mostly used when develop a project

I think the with timeout function can also change to this rule
don't override the readtime

can this be merged and make a new tag?my project need this
by the way,can we fix the tag version number?

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -685,7 +685,7 @@ func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
 		if timeout >= c.readTimeout && c.readTimeout != 0 {
 			realtimeout = c.readTimeout
 		} else if timeout <= 0 {
-			return nil, c.fatal(ErrContextCacneled)
+			return nil, c.fatal(ErrContextCanceled)
 		} else {
 			realtimeout = timeout
 		}
@@ -702,7 +702,7 @@ func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
 	}()
 	select {
 	case <-ctx.Done():
-		return nil, c.fatal(ErrContextCacneled)
+		return nil, c.fatal(ErrContextCanceled)
 	case <-endch:
 		return r, e
 	}
@@ -747,7 +747,7 @@ func (c *conn) DoContext(ctx context.Context, cmd string, args ...interface{}) (
 		if timeout >= c.readTimeout && c.readTimeout != 0 {
 			realtimeout = c.readTimeout
 		} else if timeout <= 0 {
-			return nil, c.fatal(ErrContextCacneled)
+			return nil, c.fatal(ErrContextCanceled)
 		} else {
 			realtimeout = timeout
 		}
@@ -764,7 +764,7 @@ func (c *conn) DoContext(ctx context.Context, cmd string, args ...interface{}) (
 	}()
 	select {
 	case <-ctx.Done():
-		return nil, c.fatal(ErrContextCacneled)
+		return nil, c.fatal(ErrContextCanceled)
 	case <-endch:
 		return r, e
 	}
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -677,7 +677,34 @@ func (c *conn) Flush() error {
 func (c *conn) Receive() (interface{}, error) {
 	return c.ReceiveWithTimeout(c.readTimeout)
 }
-
+func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
+	var realtimeout time.Duration
+	if dl, ok := ctx.Deadline(); ok {
+		timeout := time.Until(dl)
+		if timeout >= c.readTimeout && c.readTimeout != 0 {
+			realtimeout = c.readTimeout
+		} else if timeout <= 0 {
+			return nil, c.fatal(ErrContextCacneled)
+		} else {
+			realtimeout = timeout
+		}
+	} else {
+		realtimeout = c.readTimeout
+	}
+	endch := make(chan struct{}, 1)
+	var r interface{}
+	var e error
+	go func() {
+		r, e = c.ReceiveWithTimeout(realtimeout)
+		endch <- struct{}{}
+	}()
+	select {
+	case <-ctx.Done():
+		return nil, c.fatal(ErrContextCacneled)
+	case <-endch:
+		return r, e
+	}
+}
 func (c *conn) ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err error) {
 	var deadline time.Time
 	if timeout != 0 {
@@ -709,7 +736,34 @@ func (c *conn) ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err
 func (c *conn) Do(cmd string, args ...interface{}) (interface{}, error) {
 	return c.DoWithTimeout(c.readTimeout, cmd, args...)
 }
-
+func (c *conn) DoContext(ctx context.Context, cmd string, args ...interface{}) (interface{}, error) {
+	var realtimeout time.Duration
+	if dl, ok := ctx.Deadline(); ok {
+		timeout := time.Until(dl)
+		if timeout >= c.readTimeout && c.readTimeout != 0 {
+			realtimeout = c.readTimeout
+		} else if timeout <= 0 {
+			return nil, c.fatal(ErrContextCacneled)
+		} else {
+			realtimeout = timeout
+		}
+	} else {
+		realtimeout = c.readTimeout
+	}
+	endch := make(chan struct{}, 1)
+	var r interface{}
+	var e error
+	go func() {
+		r, e = c.DoWithTimeout(realtimeout, cmd, args)
+		endch <- struct{}{}
+	}()
+	select {
+	case <-ctx.Done():
+		return nil, c.fatal(ErrContextCacneled)
+	case <-endch:
+		return r, e
+	}
+}
 func (c *conn) DoWithTimeout(readTimeout time.Duration, cmd string, args ...interface{}) (interface{}, error) {
 	c.mu.Lock()
 	pending := c.pending
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -685,7 +685,7 @@ func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
 		if timeout >= c.readTimeout && c.readTimeout != 0 {
 			realtimeout = c.readTimeout
 		} else if timeout <= 0 {
-			return nil, c.fatal(ErrContextCanceled)
+			return nil, c.fatal(context.DeadlineExceeded)
 		} else {
 			realtimeout = timeout
 		}
@@ -702,7 +702,7 @@ func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
 	}()
 	select {
 	case <-ctx.Done():
-		return nil, c.fatal(ErrContextCanceled)
+		return nil, c.fatal(ctx.Err())
 	case <-endch:
 		return r, e
 	}
@@ -747,7 +747,7 @@ func (c *conn) DoContext(ctx context.Context, cmd string, args ...interface{}) (
 		if timeout >= c.readTimeout && c.readTimeout != 0 {
 			realtimeout = c.readTimeout
 		} else if timeout <= 0 {
-			return nil, c.fatal(ErrContextCanceled)
+			return nil, c.fatal(context.DeadlineExceeded)
 		} else {
 			realtimeout = timeout
 		}
@@ -764,7 +764,7 @@ func (c *conn) DoContext(ctx context.Context, cmd string, args ...interface{}) (
 	}()
 	select {
 	case <-ctx.Done():
-		return nil, c.fatal(ErrContextCanceled)
+		return nil, c.fatal(ctx.Err())
 	case <-endch:
 		return r, e
 	}
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -679,18 +679,18 @@ func (c *conn) Receive() (interface{}, error) {
 }
 
 func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
-	var realtimeout time.Duration
+	var realTimeout time.Duration
 	if dl, ok := ctx.Deadline(); ok {
 		timeout := time.Until(dl)
 		if timeout >= c.readTimeout && c.readTimeout != 0 {
-			realtimeout = c.readTimeout
+			realTimeout = c.readTimeout
 		} else if timeout <= 0 {
 			return nil, c.fatal(context.DeadlineExceeded)
 		} else {
-			realtimeout = timeout
+			realTimeout = timeout
 		}
 	} else {
-		realtimeout = c.readTimeout
+		realTimeout = c.readTimeout
 	}
 	endch := make(chan struct{})
 	var r interface{}
@@ -698,7 +698,7 @@ func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
 	go func() {
 		defer close(endch)
 
-		r, e = c.ReceiveWithTimeout(realtimeout)
+		r, e = c.ReceiveWithTimeout(realTimeout)
 	}()
 	select {
 	case <-ctx.Done():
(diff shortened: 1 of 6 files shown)


chenjie199234:
@stevenh pool.GetContext(ctx) only get a connection.The ctx will not effect the Do function.

```
pool := &redis.Pool{}
pool.DialCOntext = func(){
    return redis.Dial("tcp","127.0.0.1:6379")
}
func BusinessLogic(ctx context.Context) error{
    c,e:=pool.GetContext(ctx)
    //ctx didn't expired now,i can get the connection
    if e!=nil{
        return e
    }
    //.....here did some other logic
    //then the ctx is expired
    //now call the Do function
    c.Do("SET","test","abc")
    //the Do function will success.
    //because the Do function will use the readtimeout which was setted when the connection was created.but it is 0.
}
```
1.when a connection create,the connection will has 2 timeout,readtimeout and writetimeout,they are setted by DialWithReadTimeout and DialWithWriteTimeout
2.when a Do was called,the readtimeout will be used if it is not zero.
3.go the context in pool.GetContext(ctx) will only effect the connection create,it will not effect the command exec.

chenjie199234:
@stevenh can this be merged?

stevenh (changes requested):
Few things to address not mentioned in the inline comments:
* How does this deal cancelled context on connections? See https://github.com/gomodule/redigo/issues/351 for some interesting points, it's not as easy as it first appears.
* Needs tests

chenjie199234:
@stevenh
Cancel context supported.

redis/conn.go
```diff
@@ -677,7 +677,21 @@ func (c *conn) Flush() error {
 func (c *conn) Receive() (interface{}, error) {
 	return c.ReceiveWithTimeout(c.readTimeout)
 }
-
+func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
+	dl, ok := ctx.Deadline()
+	if ok {
+		timeout := time.Duration(dl.UnixNano()-time.Now().UnixNano()) * time.Nanosecond
```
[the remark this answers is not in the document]
chenjie199234: already switch to time.Until

redis/conn.go
```diff
@@ -677,7 +677,21 @@ func (c *conn) Flush() error {
...
+		if timeout >= c.readTimeout && c.readTimeout != 0 {
+			return c.ReceiveWithTimeout(c.readTimeout)
+		} else if timeout <= 0 {
+			return nil, fmt.Errorf("i/o timeout")
```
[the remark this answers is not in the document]
chenjie199234: already switch errors.New()

redis/redis.go
```diff
@@ -82,18 +85,55 @@ type Scanner interface {
...
+
+	// ReceiveWithTimeout receives a single reply from the Redis server.
+	// The timeout overrides the readtimeout set when dialing the connection.
 	ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err error)
 }
 
 var errTimeoutNotSupported = errors.New("redis: connection does not support ConnWithTimeout")
 
+// IsErrTimeout check io timeout error
+func IsErrTimeout(e error) bool {
```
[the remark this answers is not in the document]
chenjie199234: already switch to errors.New() so this function was deleted

redis/redis.go
```diff
@@ -82,18 +85,55 @@ type Scanner interface {
...
-	// Do sends a command to the server and returns the received reply.
-	// The timeout overrides the read timeout set when dialing the
-	// connection.
+	// DoContext sends a command to server and returns the received reply.
+	// This function will use the timeout in this rule:
+	// If ctx doesn't have a timeout(ctx is context.Background or context.TODO),then the readtimeout which was setted when the connection is created will be used
+	// If ctx has a timeout(ctx is created by context.WithTimeout or context.WithDeadline)
+	// If ctx's timeout is bigger then the readtimeout which was setted when the connection is created,then the readtimeout will be used
+	// If ctx's timeout is smaller then the readtimeout which was setted when the connection is created,then the ctx's timeout will be used
+	DoContext(ctx context.Context, commandName string, args ...interface{}) (reply interface{}, err error)
```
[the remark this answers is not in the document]
chenjie199234: Why this will breaks backwards compatibility.This will not effect old user's code.Do and DoWithTimeout still run in old ways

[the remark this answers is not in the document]
stevenh: You're adding a method to the `ConnWithTimeout` interface, which means anyone who has implemented this interface would need to add this to even compile with this change.

redis/conn.go
```diff
@@ -677,7 +677,34 @@ func (c *conn) Flush() error {
...
+	} else {
+		realtimeout = c.readTimeout
+	}
+	endch := make(chan struct{})
+	var r interface{}
+	var e error
+	go func() {
+		r, e = c.ReceiveWithTimeout(realtimeout)
+		endch <- struct{}{}
+	}()
```
stevenh, line 700:
```suggestion
	go func() {
	        defer close(endch)
	        
		r, e = c.ReceiveWithTimeout(realtimeout)
	}()
```
same below

redis/conn.go
```diff
@@ -709,7 +736,34 @@ func (c *conn) ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err
 func (c *conn) Do(cmd string, args ...interface{}) (interface{}, error) {
 	return c.DoWithTimeout(c.readTimeout, cmd, args...)
 }
-
+func (c *conn) DoContext(ctx context.Context, cmd string, args ...interface{}) (interface{}, error) {
```
stevenh, line 739:
```suggestion

func (c *conn) DoContext(ctx context.Context, cmd string, args ...interface{}) (interface{}, error) {
```

redis/conn.go
```diff
@@ -709,7 +736,34 @@ func (c *conn) ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err
...
+		r, e = c.DoWithTimeout(realtimeout, cmd, args)
+		endch <- struct{}{}
+	}()
+	select {
+	case <-ctx.Done():
+		return nil, c.fatal(ErrContextCacneled)
+	case <-endch:
+		return r, e
+	}
+}
```
stevenh, line 766:
```suggestion
}

```

redis/pool.go
```diff
@@ -546,7 +559,17 @@ func (ac *activeConn) Receive() (reply interface{}, err error) {
 	}
 	return pc.c.Receive()
 }
-
+func (ac *activeConn) ReceiveContext(ctx context.Context) (reply interface{}, err error) {
```
stevenh, line 562:
```suggestion

func (ac *activeConn) ReceiveContext(ctx context.Context) (reply interface{}, err error) {
```

redis/pool.go
```diff
@@ -546,7 +559,17 @@ func (ac *activeConn) Receive() (reply interface{}, err error) {
...
+	pc := ac.pc
+	if pc == nil {
+		return nil, errConnClosed
+	}
+	cwt, ok := pc.c.(ConnWithTimeout)
+	if !ok {
+		return nil, errTimeoutNotSupported
+	}
+	return cwt.ReceiveContext(ctx)
+}
```
stevenh, line 572:
```suggestion
}

```

redis/redis.go
```diff
@@ -82,18 +84,49 @@ type Scanner interface {
...
+	// The timeout overrides the readtimeout set when dialing the connection.
 	ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err error)
 }
 
 var errTimeoutNotSupported = errors.New("redis: connection does not support ConnWithTimeout")
 
+var ErrContextCacneled = errors.New("redis: context canceled")
+
+// DoContext sends a command to server and returns the received reply.
+// This function will use the timeout in this rule:
```
stevenh, line 123: I think it would be better to talk about context generically here an example of this kind of thing can be seen here:
https://golang.org/pkg/net/http/#Server.Shutdown

In addition this should be describing the behaviour not the implementation.

redis/redis.go
```diff
@@ -105,6 +138,20 @@ func DoWithTimeout(c Conn, timeout time.Duration, cmd string, args ...interface{
 	return cwt.DoWithTimeout(timeout, cmd, args...)
 }
 
+// ReceiveContext receives a single reply from the Redis server.
+// This function will use the timeout in this rule:
```
stevenh, line 148: I think it would be better to talk about context generically here an example of this kind of thing can be seen here:
https://golang.org/pkg/net/http/#Server.Shutdown

In addition this should be describing the behaviour not the implementation .

stevenh (changes requested):
No need to force push as it makes it hard to see the changes I will squash the commits to one on merge anyway. 

Here's some more feedback on top of the previous pieces.

redis/redis.go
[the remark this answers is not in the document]
chenjie199234: I see.
Already added the ```ConnWithContext``` interface

redis/redis.go
chenjie199234: take a look at the current comment.

redis/redis.go
chenjie199234: take a look at the current comment.

NitroCao:
Any progress on this? It's indeed important :(

redis/redis.go
```diff
@@ -82,17 +84,42 @@ type Scanner interface {
...
+
+	// ReceiveContext receives a single reply from the Redis server.
+	// The connection will be closed if ctx timeout or cancel when this function is running and an error "ErrContextCacneled" will return
+	ReceiveContext(ctx context.Context) (reply interface{}, err error)
+}
+
 var errTimeoutNotSupported = errors.New("redis: connection does not support ConnWithTimeout")
+var errContextNotSupported = errors.New("redis: connection does not support ConnWithContext")
+
+var ErrContextCacneled = errors.New("redis: context canceled")
```
ash2k, line 112: A typo here and where it's used
```suggestion
var ErrContextCanceled = errors.New("redis: context canceled")
```

redis/redis.go
```diff
@@ -33,6 +34,7 @@ type Conn interface {
 	Err() error
 
 	// Do sends a command to the server and returns the received reply.
+	// This function will use the timeout which was setted when the connection is created
```
ash2k, line 37:
```suggestion
	// This function will use the timeout which was set when the connection is created
```

pjebs:
Why are we redefining a brand new context cancelation error type: `var ErrContextCanceled = errors.New("redis: context canceled")`


Can't we use the one from standard library:  https://godoc.org/context#pkg-variables

Lots of people like to check the returned error and compare it to the standard library to distinguish between whether the context was cancelled or whether a deadline exceeded.

pjebs:
I think maintainer wants to go in this direction: https://github.com/gomodule/redigo/issues/351

NitroCao:
Ah, I think I should consider using github.com/go-redis/redis instead :)

stevenh (changes requested):
> **My thoughts**
> 
> 
> Why are we redefining a brand new context cancelation error type: `var ErrContextCanceled = errors.New("redis: context canceled")`
> 
> Can't we use the one from standard library: https://godoc.org/context#pkg-variables
> 
> Lots of people like to check the returned error and compare it to the standard library to distinguish between whether the context was cancelled or whether a deadline exceeded.

Good call, we should definitely do that.

> 1. This PR increases the API surface substantially (and in my opinion excessively).
> 
> All that is needed is this added to a `v2` `Conn` interface:
> 
> `func DoContext(ctx context.Context, commandName string, args ...interface{}) (reply interface{}, err error) {`
> 
> There should be no need for adding timeout/deadline facilities to make it "appear" more convenient for developers.
> Developers can easily do it themselves via the context package.
> 
> Alternatively, `Conn2` interface can be defined which embeds `Conn` interface.

Not sure I follow you there, could you clarify?

stevenh:
> Ah, I think I should consider using github.com/go-redis/redis instead :)

Is there something that is missing that is driving that comment?

chenjie199234:
@stevenh @NitroCao @pjebs @ash2k

chenjie199234:
@stevenh @NitroCao @pjebs @ash2k 
sorry guys,work and life have some change,i will find some time to fix the comments,but it may have some delay

chenjie199234:
@pjebs @stevenh already changed context error to std lib's error.
test file fixed too

redis/redis.go
```diff
@@ -82,17 +84,42 @@ type Scanner interface {
...
+
+	// ReceiveContext receives a single reply from the Redis server.
+	// The connection will be closed if ctx timeout or cancel when this function is running and an error "ErrContextCacneled" will return
+	ReceiveContext(ctx context.Context) (reply interface{}, err error)
+}
+
 var errTimeoutNotSupported = errors.New("redis: connection does not support ConnWithTimeout")
+var errContextNotSupported = errors.New("redis: connection does not support ConnWithContext")
+
+//var ErrContextCanceled = errors.New("redis: context canceled")
```
stevenh, line 112: Old commented out code

redis/redis.go
```diff
@@ -82,17 +84,42 @@ type Scanner interface {
...
+	// The timeout overrides the readtimeout set when dialing the connection.
 	ReceiveWithTimeout(timeout time.Duration) (reply interface{}, err error)
 }
 
+// ConnWithContext is an optional interface that allows the caller to control the command's life with context
+type ConnWithContext interface {
+	Conn
+
+	// DoContext sends a command to server and returns the received reply.
+	// The connection will be closed if ctx timeout or cancel when this function is running and an error "ErrContextCacneled" will return
```
stevenh, line 101: Comment looks out of date as it refers to `ErrContextCacneled`

redis/redis.go
```diff
@@ -82,17 +84,42 @@ type Scanner interface {
...
+	DoContext(ctx context.Context, commandName string, args ...interface{}) (reply interface{}, err error)
+
+	// ReceiveContext receives a single reply from the Redis server.
+	// The connection will be closed if ctx timeout or cancel when this function is running and an error "ErrContextCacneled" will return
```
stevenh, line 105: Comment looks out of date as it refers to `ErrContextCacneled`

redis/redis.go
```diff
@@ -82,17 +84,42 @@ type Scanner interface {
...
+	ReceiveContext(ctx context.Context) (reply interface{}, err error)
+}
+
 var errTimeoutNotSupported = errors.New("redis: connection does not support ConnWithTimeout")
+var errContextNotSupported = errors.New("redis: connection does not support ConnWithContext")
+
+//var ErrContextCanceled = errors.New("redis: context canceled")
+
+// DoContext sends a command to server and returns the received reply.
+// The connection will be closed if ctx timeout or cancel when this function is running and an error "ErrContextCacneled" will return
```
stevenh, line 115: Comment looks out of date as it refers to `ErrContextCacneled`

redis/redis.go
```diff
@@ -105,6 +132,16 @@ func DoWithTimeout(c Conn, timeout time.Duration, cmd string, args ...interface{
 	return cwt.DoWithTimeout(timeout, cmd, args...)
 }
 
+// ReceiveContext receives a single reply from the Redis server.
+// The connection will be closed if ctx timeout or cancel when this function is running and an error "ErrContextCacneled" will return
```
stevenh, line 136: Comment looks out of date as it refers to `ErrContextCacneled`

redis/redis_test.go
```diff
@@ -1,4 +1,5 @@
 // Copyright 2017 Gary Burd
+
```
stevenh, line 2:
```suggestion
```

stevenh (changes requested):

redis/conn.go
```diff
@@ -678,6 +678,36 @@ func (c *conn) Receive() (interface{}, error) {
 	return c.ReceiveWithTimeout(c.readTimeout)
 }
 
+func (c *conn) ReceiveContext(ctx context.Context) (interface{}, error) {
+	var realtimeout time.Duration
```
nicholas-eden, line 682:
```suggestion
	var realTimeout time.Duration
```

TheFireMike:
Any updates on this? We would really like to use this library in our [Thola](https://github.com/inexio/thola/) project, but Context support is critical for us.

stevenh:
@TheFireMike this is now merged, sorry I'm not been getting some github emails, so didn't realise this was now complete.

Thanks to everyone for their contributions to this PR.

KebSeb:
may i ask, why i cant find DoContext in the current realease (v1.8.5)?

stevenh:
It was merged after [v1.8.5](https://github.com/gomodule/redigo/releases/tag/v1.8.5).

I've just created [v1.8.6](https://github.com/gomodule/redigo/releases/tag/v1.8.6)

KebSeb:
this is awesome, thank you very much

PR #551, opened 2021-02-01
Remove unnecessary blank (_) identifier

withshubh:
Hi :wave: 
I ran the [DeepSource](https://deepsource.io) static analyzer on the forked copy of this repo and found some [interesting code quality issues](https://deepsource.io/gh/withshubh/redigo/issues/). This PR fixed a few of them.

#### Summary of changes
- Remove unnecessary blank (_) identifier
- ~Added .deepsource.toml~


.deepsource.toml
```diff
@@ -0,0 +1,10 @@
+version = 1
```
stevenh, line 1: Looks like a local file which shouldn't be included?

stevenh (changes requested):

.deepsource.toml
withshubh: This is a configuration file required to activate analysis in the repo. Apart from finding [issues](https://deepsource.io/gh/withshubh/redigo/issues/?category=recommended) in Go, You can use DeepSource to track test coverage, detect problems in Dockerfiles.

withshubh: @stevenh any update on this? :sparkling_heart:

stevenh: Hi @wricardo sorry I missed you reply there.

I'm happy merge the code fix but would need to do some more research on deepsource as looks like its triggering some false positives in that link.

For other projects I've used [golangci-lint](https://github.com/golangci/golangci-lint) which is good so will have a look at adding that here too.

stevenh: Draft of the golangci-lint is here: https://github.com/gomodule/redigo/pull/554 if you're interested in helping review as there are lots of fixes.

withshubh: @stevenh The false-positive rate of the DeepSource analyzer is less than 5%. You can integrate it to track test coverage, Detect problems in Dockerfiles, etc. in addition to detecting issues in Go. Read more about it [here](https://deepsource.io/go-static-code-analysis/).

stevenh: 5% is pretty high IMO, lets get the fix merged without deepsource then we can look at that later, without blocking the fix.

withshubh:
@stevenh I have removed the configuration file.

I would really appreciate it if you could drop a suggestion on what we can do onboard you(or redigo) on DeepSource in the future. :sparkling_heart:

withshubh:
@stevenh :wave: just want to follow up on this!

stevenh:
Thank you for the PR, [new release](https://github.com/gomodule/redigo/releases/tag/v1.8.4) was cut last night too, which includes a bunch more fixes :)

withshubh:
Can I submit a new PR with the configuration file which I used to configure the analysis?
You can merge the configuration file to activate DeepSource and fix [these issues](https://deepsource.io/gh/withshubh/redigo/issues/?category=bug-risk).

stevenh:
I believe all of those are already fixed.

If there are issues which DeepSource detects that aren't I'd like to hear about that.

withshubh:
I updated my forked master and ran the analysis. DeepSource caught [security issues](https://deepsource.io/gh/withshubh/redigo/issues/?category=security) that you might want to fix.

stevenh:
Thanks @withshubh.

I had a quick look, the sha1 errors are false positives as its not being used for crypto and the tls one is a fallback when the user doesn't provide a config, hence needs to be permissive for usability. It could be argued that this should just return an error but that wouldn't be backwards compatible :(

withshubh:
Got it.
Thanks! @stevenh :sparkles:

PR #554, opened 2021-02-07
chore: Enable golangci-lint and fix errors

stevenh:
Enable golangci-lint checking in travis and fix errors identified by it.

### DIFF
--- a/redis/pubsub.go
+++ b/redis/pubsub.go
@@ -60,27 +60,35 @@ func (c PubSubConn) Close() error {
 
 // Subscribe subscribes the connection to the specified channels.
 func (c PubSubConn) Subscribe(channel ...interface{}) error {
-	c.Conn.Send("SUBSCRIBE", channel...)
+	if err := c.Conn.Send("SUBSCRIBE", channel...); err != nil {
+		return err
+	}
 	return c.Conn.Flush()
 }
 
 // PSubscribe subscribes the connection to the given patterns.
 func (c PubSubConn) PSubscribe(channel ...interface{}) error {
-	c.Conn.Send("PSUBSCRIBE", channel...)
+	if err := c.Conn.Send("PSUBSCRIBE", channel...); err != nil {
+		return err
+	}
 	return c.Conn.Flush()
 }
 
 // Unsubscribe unsubscribes the connection from the given channels, or from all
 // of them if none is given.
 func (c PubSubConn) Unsubscribe(channel ...interface{}) error {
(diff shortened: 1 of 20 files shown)


stevenh:
@darkliquid @danmrichards could I get a review on this please

redis/pubsub.go
```diff
@@ -60,27 +60,35 @@ func (c PubSubConn) Close() error {
 
 // Subscribe subscribes the connection to the specified channels.
 func (c PubSubConn) Subscribe(channel ...interface{}) error {
-	c.Conn.Send("SUBSCRIBE", channel...)
+	if err := c.Conn.Send("SUBSCRIBE", channel...); err != nil {
+		return err
+	}
```
darkliquid, line 65: Changes like this seem quite dangerous. While this is the _correct_ behaviour, I do worry about the effect on applications using this package depending on the current behaviour of flush the connection regardless of the outcome of the send. Potentially this warrants a minor or even major version bump depending on how impactful this could be.

redis/pubsub_example_test.go
```diff
@@ -116,9 +118,18 @@ func publish() {
 	}
 	defer c.Close()
 
-	c.Do("PUBLISH", "c1", "hello")
-	c.Do("PUBLISH", "c2", "world")
-	c.Do("PUBLISH", "c1", "goodbye")
+	if _, err = c.Do("PUBLISH", "c1", "hello"); err != nil {
+		fmt.Println(err)
```
darkliquid, line 122: is there a reason for using `fmt.Println` here over say `t.Log()`?

stevenh: It's because this is an example not a test, so trying to be more real.

redis/pubsub.go
stevenh: My take on **this particular case** is that the `Flush` should be returning an error if the `Send` failed i.e. if `Send` failed `Flush` should also. Thoughts?

darkliquid: Yeah, digging into it a bit more, conn is using a `bufio.Writer` for writes, whose behaviour described here https://golang.org/pkg/bufio/#Writer indicates that should a `Write` call to it fail (which is what `Send` will ultimately end up doing) then subsequent calls to `Flush` should also fail, so yeah, I think it's fine as it shouldn't actually result in a change in behaviour anyway.

darkliquid (approved):

danmrichards (approved):
LGTM

Issue #375, opened 2018-11-20
Redis streams


samwhitecoull:
I am using Redigo in a project that relies on reading messages from streams. I know this is a newer feature of Redis, but is this something that Redigo plans to support natively in the future? I have cobbled together a stream reader by traversing through the nested interfaces returned from the `Do()` call and using type assertion (based on trial and improvement and knowledge of whats being sent) to pull useful results out. However this is extremely brittle and only works because I can guarantee whats being appended to the streams. If a code update is not required it would be great to see some documentation showing an idiomatic way of reading streams.

smartwalle:
r, _ := redis.Values(conn.Do("XREAD", "STREAMS", "stream_key", "0-0"))
	for kIndex :=0; kIndex < len(r); kIndex++ {
		var keyInfo = r[kIndex].([]interface{})

		var key = string(keyInfo[0].([]byte))
		var idList = keyInfo[1].([]interface{})

		for idIndex :=0; idIndex <len(idList); idIndex++ {
			var idInfo = idList[0].([]interface{})

			var id = string(idInfo[0].([]byte))

			var fieldList = idInfo[1].([]interface{})
			var field = string(fieldList[0].([]byte))
			var value = string(fieldList[1].([]byte))

			fmt.Println(key, id, field, value)
		}
	}

smartwalle:
```
r, _ := redis.Values(conn.Do("XREAD", "STREAMS", "stream_key", "0-0"))
	for kIndex :=0; kIndex < len(r); kIndex++ {
		var keyInfo = r[kIndex].([]interface{})

		var key = string(keyInfo[0].([]byte))
		var idList = keyInfo[1].([]interface{})

		for idIndex :=0; idIndex <len(idList); idIndex++ {
			var idInfo = idList[0].([]interface{})

			var id = string(idInfo[0].([]byte))

			var fieldList = idInfo[1].([]interface{})
			var field = string(fieldList[0].([]byte))
			var value = string(fieldList[1].([]byte))

			fmt.Println(key, id, field, value)
		}
	}
```

smartwalle:
```
r, err := redis.Values(conn.Do("XREAD", "STREAMS", "stream1", "stream2", "0-0", "0-0"))

for kIndex :=0; kIndex < len(r); kIndex++ {
	var keyInfo = r[kIndex].([]interface{})

	var key = string(keyInfo[0].([]byte))
	var idList = keyInfo[1].([]interface{})

[...]

PR #557, opened 2021-03-24
add parsing for stream entries

smotes:
This PR is related to issue https://github.com/gomodule/redigo/issues/375 and adds a first pass at parsing replies containing stream entries into a struct.

I tested things against a Redis server v5.0.7, which is the first version containing the streams feature: [https://redislabs.com/blog/redis-5-0-is-here/](https://redislabs.com/blog/redis-5-0-is-here/).

### DIFF
--- a/redis/reply_test.go
+++ b/redis/reply_test.go
@@ -143,6 +143,27 @@ var replyTests = []struct {
 		ve(getSlowLog()),
 		ve(redis.SlowLog{ID: 1, Time: time.Unix(1579625870, 0), ExecutionTime: time.Duration(3) * time.Microsecond, Args: []string{"set", "x", "y"}, ClientAddr: "localhost:1234", ClientName: "testClient"}, nil),
 	},
+	{
+		"[Entry(0-1, name, Anna), Entry(1-1, name, Bruce)]",
+		ve(redis.Entries([]interface{}{
+			[]interface{}{[]byte("0-1"), []interface{}{[]byte("name"), []byte("Anna")}},
+			[]interface{}{[]byte("1-1"), []interface{}{[]byte("name"), []byte("Bruce")}},
+		}, nil)),
+		ve([]redis.Entry{
+			{
+				ID: "0-1",
+				Fields: map[string]string{
+					"name": "Anna",
+				},
+			},
+			{
+				ID: "1-1",
+				Fields: map[string]string{
+					"name": "Bruce",
+				},
+			},
+		}, nil),
+	},
 }
 
 func getSlowLog() (redis.SlowLog, error) {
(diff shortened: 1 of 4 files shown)


redis/reply_test.go
```diff
@@ -220,6 +241,58 @@ func TestSlowLog(t *testing.T) {
...
+		}
+	}
+
+	entries, err := redis.Entries(c.Do("XRANGE", "teststream", "-", "+"))
+	if err != nil {
+		t.Errorf("TestEntries failed during XRANGE with error " + err.Error())
+		return
+	}
+	if len(entries) != n {
+		t.Errorf("TestEntries expected %d entries in result, got %d", n, len(entries))
```
stevenh, line 277:
```suggestion
		t.Errorf("TestEntries expected %d entries in result, got %d", n, len(entries))
		return
```

redis/reply_test.go
```diff
@@ -220,6 +241,58 @@ func TestSlowLog(t *testing.T) {
 	}
 }
 
+func TestEntries(t *testing.T) {
+	c, err := dial()
+	if err != nil {
+		t.Errorf("TestEntries failed during dial with error " + err.Error())
```
stevenh, line 247: Let's not name the test as we get the context from the test out.

redis/reply_test.go
```diff
@@ -220,6 +241,58 @@ func TestSlowLog(t *testing.T) {
...
+	defer c.Close()
+
+	resultStr, err := redis.Strings(c.Do("CONFIG", "GET", "stream-node-max-entries"))
+	if err != nil {
+		t.Errorf("TestEntries failed during CONFIG GET stream-node-max-entries with error " + err.Error())
+		return
+	}
+	// in case of older verion < 5.0 where streams are not supported don't run the test
+	if len(resultStr) == 0 {
+		return
```
stevenh, line 259: We should use skip here to make it explicit

stevenh (changes requested):
Thanks for looking at this.

I've done a quick pass and adding some initial feedback

smotes:
> 
> 
> Thanks for looking at this.
> 
> I've done a quick pass and adding some initial feedback

See commit 4a86003844a18575e980d5069d371d3e541bc74a.

tk42:
Is this working for XREAD or XREADGROUP? I wonder this is not working well for them because the nested structure is different, sorry if this is out of scope.

stevenh (approved):

stevenh:
> Is this working for `XREAD` or `XREADGROUP`? I wonder this is not working well for them because the nested structure is different, sorry if this is out of scope.

You raise a valid question @tk42 it does look like this will only work for `XRANGE` which I don't think is clear from the definition of Entries. So I wonder if there is a better way to do this which is more generic.

Any thoughts @smotes ?

smotes:
Hi @stevenh.

I propose we keep things as they stand. The Redis docs distinguish between entries and stream names in return types, so I felt it on the caller to parse out the reply in parts and where to find the nested entries within.

Consider the `XREADGROUP` Redis command docs (https://redis.io/commands/xreadgroup#return-value).

> Array reply, specifically:
> The command returns an array of results: each element of the returned array is an array composed of a two element containing the key name and the entries reported for that key. The entries reported are full stream entries, having IDs and the list of all the fields and values. Field and values are guaranteed to be reported in the same order they were added by XADD.

And here's a little code snippet showing how to parse that as is (adapted from some code in a personal project).

```go
reply, err := sc.conn.Do("XREADGROUP", "GROUP", "some-group-name", "some-consumer-name", 
	"BLOCK", 3000, "COUNT", 10, "STREAMS", "some-stream-name", ">")
	
vs, err := redis.Values(reply, err)
if err != nil {
	// Ignore the error on empty reads, wait until the next.
	if errors.Is(err, redis.ErrNil) {
		continue
	}
	return err
}

// Get the first and only value in the array since we're only
// reading from one stream here.
vs, err = redis.Values(vs[0], nil)
if err != nil {
	return err
}

// Ignore the stream name as the first value as we already have that in hand!
// Just get the second value which is guaranteed to exist, and parse it as stream entries.
entries, err := ParseEntries(vs[1], nil)
if err != nil {
	return fmt.Errorf("error parsing entries: %w", err)
}
```

stevenh:
That makes sense to me @tk42 does that answer address your concerns?

tk42:
Hi @smotes @stevenh
In that snippet, it contains only a single stream key but how about multiple stream key case? From redis docs, ```XREAD```/```XREADGROUP``` can take multiple keys (and corresponded ids).

> XREAD [COUNT count] [BLOCK milliseconds] STREAMS key **[key ...]** id **[id ...]**
> XREADGROUP GROUP group consumer [COUNT count] [BLOCK milliseconds] [NOACK] STREAMS key **[key ...]** id **[id ...]**

I’m feeling that users for ```XREAD```/```XREADGROUP``` will call ```redis.Values``` multiple times as a template pattern, so a new struct of return-type for them  might help those users. Something like this ;)
```golang
type StreamEntry struct {
    Stream       string
    Entries      []Entry
}
```
and its corresponded parsing function like this
```golang
func StreamEntries(reply interface{}, err error) ([]StreamEntry, error) {
    vss, err := redis.Values(reply, err)
    if err != nil {
        // error handling
	    return nil, fmt.Errorf("error parsing reply: %w", err)
    }

    var streamEntries []StreamEntry
    for _, vs := range vss {
        name, err := redis.String(vs[0], nil)
        if err != nil {
            // error handling
            streamEntries = append(streamEntries, new(StreamEntry))
        }

        entries, err := ParseEntries(vs[1], nil)
        if err != nil {
            // error handling
            streamEntries = append(streamEntries, new(StreamEntry))
        }

        streamEntries = append(streamEntries, StreamEntry{
               Stream: name,
               Entries: entries,
        })
    }
    return streamEntries, nil
}
```

PR #573, opened 2021-09-24
feat: add DialCtx DialOption

### LINKED ISSUE #572: IsDialURLContext useful?
I needed a DialURLContext(), and so wrote one. (not hard)

Do the maintainers see this as useful for the project?

PS, Thanks for creating and maintaining Redigo

stevenh:
Add DialCtx which is a DialOption which specifies the context to use when performing dial requests.

This also deprecates DialContext, which in hindsight should have been implemented this way as it allows all Dial methods to support context.

Fixes #572


stevenh:
Abandoning

PR #574, opened 2021-09-26
Implement DialURLContext(...)

cameronelliott:
and convert DialURL to call DialURLContext()

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -316,9 +316,7 @@ func DialContext(ctx context.Context, network, address string, options ...DialOp
 
 var pathDBRegexp = regexp.MustCompile(`/(\d*)\z`)
 
-// DialURL connects to a Redis server at the given URL using the Redis
-// URI scheme. URLs should follow the draft IANA specification for the
-// scheme (https://www.iana.org/assignments/uri-schemes/prov/redis).
+// DialURL wraps DialURLContext using context.Background.
 func DialURL(rawurl string, options ...DialOption) (Conn, error) {
 	ctx := context.Background()
 
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -320,6 +320,15 @@ var pathDBRegexp = regexp.MustCompile(`/(\d*)\z`)
 // URI scheme. URLs should follow the draft IANA specification for the
 // scheme (https://www.iana.org/assignments/uri-schemes/prov/redis).
 func DialURL(rawurl string, options ...DialOption) (Conn, error) {
+	ctx := context.Background()
+
+	return DialURLContext(ctx, rawurl, options...)
+}
+
+// DialURLContext connects to a Redis server at the given URL using the Redis
+// URI scheme. URLs should follow the draft IANA specification for the
+// scheme (https://www.iana.org/assignments/uri-schemes/prov/redis).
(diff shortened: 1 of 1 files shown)


cameronelliott:
Hey @stevenh this is what I was thinking, although maybe there are reasons to pass the 
context as an option, and it certainly would work functionally.

But it might confuse people.

I have never seen a context.Context passed as an option like that, and the Google
official manner is or was:

> At Google, we require that Go programmers pass a Context parameter as the first argument to every function on the call path between incoming and outgoing requests. 
Based upon this article:
https://go.dev/blog/context

And in my experience, the idiomatic Go way is to have a function named xxxContext when
you want to pass a context.Context in.

So, I might miss there is a way to pass a context.Context in if I didn't see
the method in the method signatures in the index, like in the typical Dial*Context method names.

But maybe it makes sense if the Dial method signatures are multiplying too much.

redis/conn.go
```diff
@@ -320,6 +320,15 @@ var pathDBRegexp = regexp.MustCompile(`/(\d*)\z`)
 // URI scheme. URLs should follow the draft IANA specification for the
 // scheme (https://www.iana.org/assignments/uri-schemes/prov/redis).
 func DialURL(rawurl string, options ...DialOption) (Conn, error) {
+	ctx := context.Background()
+
+	return DialURLContext(ctx, rawurl, options...)
```
stevenh, line 325:
```suggestion
	return DialURLContext(context.Background(), rawurl, options...)
```

redis/conn.go
```diff
@@ -320,6 +320,15 @@ var pathDBRegexp = regexp.MustCompile(`/(\d*)\z`)
 // URI scheme. URLs should follow the draft IANA specification for the
 // scheme (https://www.iana.org/assignments/uri-schemes/prov/redis).
```
stevenh, line 321: I suggest we update the docs for `DialURL` to
```golang
// DialURL wraps DialURLContext using context.Background.
```

stevenh:
Feels like we should change all Dialers to use ctx as the first arg in v2.

I believe xxxContent function naming is due to the lack of context support in early go versions + v1 compatibility guarantee, so the original functions must be kept.

Consistency vs other APIs which require ctx as first parameter is good point, it has the benefit that it enforces the design that a consumer should pass a valid context vs making less obvious with an additional functional option.

stevenh (changes requested):
Thanks for the PR, I agree this is a better approach.

Lets get the docs for older none ctx Dial functions all updated to explicitly call out their wrap nature.

cameronelliott:
Can you suggest/write the Conventional Commit labels?
I am not familiar with it.

stevenh:
> Can you suggest/write the Conventional Commit labels? I am not familiar with it.

Sure It's really simple this is a feature so you would have something like this:
> feat: add DialURLContext
>
> Add DialURLContext so that we can consumers have
> control over cancelation and timeout

redis/conn.go
stevenh: Just missing this @cameronelliott

stevenh (changes requested):
Thanks for the changes, just missing the suggested doc change for DialURL which makes it clear to the caller its a wrapper.

cameronelliott:
@stevenh Thanks for taking this PR, and your patience while I got it into some type of passable shape.

PR #577, opened 2021-10-27
add a simple example with 3 function 'GetValue' 'SetValue' 'GetKeys'

oneleg:


oneleg:
Hi yes I saw the test sample page but I thought it would be easier for others to have a file with  functions 😬
anyway thanks for response to my request 👍🏻🌹

oneleg:
> Thanks for the PR @oneleg but not sure what the benefit of this is as it doesn't use a typical flow, specifically leveraging a connection pool and is already pretty well covered by [docs](https://pkg.go.dev/github.com/gomodule/redigo/redis).
> 
> A better way to have concrete examples would via [test examples](https://pkg.go.dev/testing#hdr-Examples).

Hi yes I saw the test sample page but I thought it would be easier for others to have a file with functions grimacing
anyway thanks for response to my request 👍🏻rose

stevenh:
Happy to reconsider if it’s updated to use best practices including a pool, but for now closing

PR #589, opened 2021-12-24
add pool FILO mode

tangtj:


tangtj:
https://github.com/gomodule/redigo/pull/571

stevenh:
There is already a https://github.com/gomodule/redigo/pull/571 for this. The same issue applies to this is that this breaks connection cleanup and is only needed to solve a problem outside of this package

tangtj:
thanks you.
We consulted the technical staff of the cloud vendor, and the solution given is this method. I am willing to participate in the follow-up and solve the related problems.

stevenh:
Happy to consider options but from what I’ve read about the situation it would need a more situation aware solution to fix properly.

stevenh:
I wonder if the best fix for this would be to limit the connection lifetime, which would force rebalance over time?

tangtj:
If the lifetime is very short, Pool will be created and closed conn frequently. The use of pool is meaningless and is equivalent to creating a new conn every time. If the time is longer, for example, 20 seconds or 1 minute，Then the rebalan is has no effect，unbalanced state transfers on different connection，single unbalanced connection will still become the overall performance bottleneck。

stevenh:
That would only be the case if the LB isn't doing it's job properly, and if that is the case no fix will work.

From what I understood the balance issue only occurred where there was an issue, hence a 1min max connection lifetime would enable it to only be unbalanced after said issue for 1mins, which would maintain the full benefit of the pool?

PR #594, opened 2022-02-01
issue_592 fixed

mohammadrezayousefpour:


stevenh (changes requested):
There isn't an issue 594 and this change just renames a function without adjusting the code a `ReceiveContext` would need to take a `context.Context` not a `time.Duration`

stevenh:
Going to close this as code is unfortunately invalid

PR #595, opened 2022-02-01
Issue_351 fixed

mohammadrezayousefpour:


redis/conn.go
```diff
@@ -426,9 +427,13 @@ func (c *conn) fatal(err error) error {
 }
 
 func (c *conn) Err() error {
-	c.mu.Lock()
-	err := c.err
-	c.mu.Unlock()
+	if(c.err !=nil)
```
stevenh, line 430: Accessing without the lock is a data race.

redis/conn.go
```diff
@@ -414,8 +414,9 @@ func (c *conn) Close() error {
 }
 
 func (c *conn) fatal(err error) error {
-	c.mu.Lock()
+	
 	if c.err == nil {
+		c.mu.Lock()
```
stevenh, line 419: As below this will cause a data race.

stevenh (changes requested):
Unfortunately all the changes here are invalid and cause data races and they also don't seem to reflect what #351 is referring too.

stevenh:
Due to this being invalid code going to close this.

PR #596, opened 2022-02-01
Scan Struct(issue_570) fixed

mohammadrezayousefpour:


redis/scan.go
```diff
@@ -27,12 +27,40 @@ var (
...
+        case []byte:
+                return t.UnmarshalText(src)
+        default:
+                return fmt.Errorf("cannot convert from %T to %T", src, t)
+        }
+}
+
 func ensureLen(d reflect.Value, n int) {
-	if n > d.Cap() {
+	if(n!=nil && d!=nil)
```
stevenh, line 55: Unfortunately This in invalid code, `n` is an `int` so can't be nil as such it doesn't compile as can be seen here:
https://go.dev/play/p/H4HgPUpzxPY

redis/scan.go
```diff
@@ -27,12 +27,40 @@ var (
...
+
+        switch src := src.(type) {
+        case string:
+                return t.UnmarshalText([]byte(src))
+        case []byte:
+                return t.UnmarshalText(src)
+        default:
+                return fmt.Errorf("cannot convert from %T to %T", src, t)
+        }
+}
```
stevenh, line 52: This code is never used, so not sure why it's needed?

stevenh (changes requested):

stevenh:
Thanks for the PR's @mohammadrezayousefpour but unfortunately they all have some serious issues making them invalid.

Happy to discuss fixes on the relevant issues and for you to raise new PR's, but please refer to [this PR as an example](https://github.com/gomodule/redigo/pull/580) of how we would like to see that done.

Issue #592, opened 2022-01-16
Add PubSubConn.ReceiveContext()


RelicOfTesla:
1. add "PubSubConn.ReceiveContext" support watch the <-context.Done() .
 because "redis.ReceiveContext" is return native redis response. was not call "Subscription.receiveInternal" decode this response.

2. and, "ReceiveWithTimeout" is return net.ErrClosed, can not loop call receive

stevenh:
I'm sorry I don't understand what you're trying to say, do you have some code which demonstrates the issue you have?

RelicOfTesla:
1.
```go

func run(ctx context.Context){
  psc := &redis.PubSubConn{Conn: conn}
  err := psc.Subscribe(topicStr)
  for {
    r: = psc.ReceiveContext(ctx)   // add this
    err, ok := r.(error)
    _ = ok
    if errors.Is(v, context.DeadlineExceeded) || errors.Is(v, context.Canceled) {
        break; // break loop
    }
    msg, ok := r.(redis.Message)
    // ....

    // if not have the psc.ReceiveContext(ctx), and i try "v, err := redis.ReceiveContext(psc.Conn, ctx)"
    // the v was can not convert redis.Message, because not decode by Subscription.receiveInternal
    // so , please add PubSubConn.ReceiveContext
  }
}
func main(){
   ctx,cancel := context.WithCancel(context.TODO())
   go run(ctx)
   time.Sleep(time.Second*10)
    cancel()
   time.Sleep(time.Second*10)
}
```


2.
```go
for {
   v := psc.ReceiveWithTimeout(time.Second)
  fmt.Println(v) 
  // first print: os.ErrDeadlineExceeded,  and this connection was closed!!!!
  // second print: net.ErrClosed  
}


```

stevenh:
To confirm my understanding you're looking for a new feature `PubSubConn.ReceiveContext(ctx context.Context)`?

You're also saying that `psc.ReceiveWithTimeout` returns an different error indicated the connection is closed after a timeout occurs. This is indeed the case, and is by design, and that would also be the same for a `Context` method as a timeout is treated as a fatal error, which can be seen [here](https://github.com/gomodule/redigo/blob/574218b5788d47ef16ad8eb15a6704314a3e42d5/redis/conn.go#L737). It's a fatal error as with you can't be sure what state the connection is in, so it must be closed.

RelicOfTesla:
These are two questions.....   second question it was an unexpected discovery

[...]

PR #603, opened 2022-03-17
feat: add RequestContext to PubSubConn

pabigot:
Add a wrapper that goes through the standard receiveInternal
processing to match the API of the existing PubSubConn Receive
methods.

Fixes: #592

### DIFF
--- a/redis/pubsub.go
+++ b/redis/pubsub.go
@@ -15,6 +15,7 @@
 package redis
 
 import (
+	"context"
 	"errors"
 	"time"
 )
@@ -114,6 +115,13 @@ func (c PubSubConn) Receive() interface{} {
 // override the connection's default timeout.
 func (c PubSubConn) ReceiveWithTimeout(timeout time.Duration) interface{} {
 	return c.receiveInternal(ReceiveWithTimeout(c.Conn, timeout))
+}
+
+// ReceiveContext is like Receive, but it allows termination of the receive
+// via a Context. If the call returns due to closure of the context's Done
+// channel the underlying Conn will have been closed.
+func (c PubSubConn) ReceiveContext(ctx context.Context) interface{} {
+	return c.receiveInternal(ReceiveContext(c.Conn, ctx))
 }
 
 func (c PubSubConn) receiveInternal(replyArg interface{}, errArg error) interface{} {
--- a/redis/pubsub_test.go
+++ b/redis/pubsub_test.go
@@ -15,6 +15,8 @@
 package redis_test
 
 import (
+	"context"
+	"errors"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


pabigot:
It doesn't seem that the design intent is to limit the features of PubSubConn, but an alternative approach that's only slightly less ideal would be to expose `pubsub/receiveInternal` as a visible, documented reply helper so we could do:

    m, err := redis.PubSubReply(redis.ReceiveContext(c, ctx))

That might be useful anyway.

redis/pubsub.go
```diff
@@ -116,6 +117,12 @@ func (c PubSubConn) ReceiveWithTimeout(timeout time.Duration) interface{} {
 	return c.receiveInternal(ReceiveWithTimeout(c.Conn, timeout))
 }
 
+// ReceiveContext is like Receive, but it allows termination of the receive
+// via a Context.
```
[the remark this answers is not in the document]
stevenh: 👍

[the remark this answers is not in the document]
pabigot: @stevenh I can't tell what the thumbs-up indicates--that you agree it needed documentation, or that the documentation I subsequently added was acceptable, or....?  If you have any changes you require please let me know; from my perspective this is ready.

redis/pubsub.go
```diff
@@ -116,6 +117,13 @@ func (c PubSubConn) ReceiveWithTimeout(timeout time.Duration) interface{} {
 	return c.receiveInternal(ReceiveWithTimeout(c.Conn, timeout))
 }
 
+// ReceiveContext is like Receive, but it allows termination of the receive
+// via a Context.  If the call returns due to closure of the context's Done
```
stevenh, line 121:
```suggestion
// via a Context. If the call returns due to closure of the context's Done
```

redis/pubsub_test.go
```diff
@@ -74,3 +76,25 @@ func TestPushed(t *testing.T) {
...
+
+	c := redis.PubSubConn{Conn: sc}
+
+	require.NoError(t, c.Subscribe("c1"))
+	expectPushed(t, c, "Subscribe(c1)", redis.Subscription{Kind: "subscribe", Channel: "c1", Count: 1})
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	got := c.ReceiveContext(ctx)
+	if err := got.(error); !errors.Is(err, context.Canceled) {
```
stevenh, line 95: If this doesn't return an error it will panic, which I don't believe is the intention?

stevenh (changes requested):

redis/pubsub_test.go
pabigot: Added checks for this.

redis/pubsub.go
pabigot: done

stevenh (approved):

stevenh:
Thanks for the PR

PR #604, opened 2022-03-17
chore: retract redis and redisx sub modules

stevenh:
Retract the non versioned entries of redis and redisx modules to fix pkg.go.dev versioning.

### DIFF
--- a/redis/go.mod
+++ b/redis/go.mod
@@ -1,0 +1,7 @@
+module github.com/gomodule/redigo/redis
+
+go 1.16
+
+require github.com/stretchr/testify v1.7.0
+
+retract [v0.0.0-0, v0.0.1] // Never used only present to fix pkg.go.dev.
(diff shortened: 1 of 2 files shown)


redis/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/gomodule/redigo/redis
+
+go 1.16
+
+require github.com/stretchr/testify v1.7.0
+
+retract (
+	v0.0.1 // Never used only present to fix pkg.go.dev.
+	v0.0.0-0 // Never used only present to fix pkg.go.dev.
```
seankhliao, line 9: the syntax you want is a range, not 2 individual versions: `[v0.0.0-0, v0.0.1]`

pabigot:
I like this approach.  An enhancement suggested by https://github.com/golang/go/issues/51720#issuecomment-1071497694 would be to not merge this at all.   Make a v0.0.1 tag pointing to the head of your PR branch (or push directly to a non-main branch here), push the tag, do the proxy fetch that will trigger the retraction, then delete the tag.  No need to clutter the main branch with this stuff.

Assuming I understand how this magic works.

stevenh:
Merge not needed

PR #612, opened 2022-05-31
Fixing error message for SlowLog conversion

dmitri-lerko:
Hello :wave:

I've noticed that error handling for SlowLogs parser is incorrectly references element 0. This should fix it.

Thanks


stevenh:
@dmitri-lerko could you rebase to the latest master to pickup the golangci-lint fix that if causing failures on golang 1.18 please.

PR #614, opened 2022-06-04
Support LATENCY LATEST, LATEST HISTORY command parsing

dmitri-lerko:
Includes tests similar to how SlowLog was tested.

### DIFF
--- a/redis/redis.go
+++ b/redis/redis.go
@@ -187,3 +187,27 @@ type SlowLog struct {
 	// ClientName is the name set via the CLIENT SETNAME command (4.0 only).
 	ClientName string
 }
+
+// Latency represents a redis LATENCY LATEST
+type Latency struct {
+	// Latency event name
+	Event string
+
+	// Time is the unix timestamp at which the event was processed.
+	Time time.Time
+
+	// LatestExecutationTime is the amount of time needed for the command execution.
+	LatestExecutionTime time.Duration
+
+	// Records the all-time maximum latency for every element
+	MaxExecutionTime time.Duration
+}
+
+// LatencyHistory represents a redis LATENCY HISTORY
+type LatencyHistory struct {
+	// Time is the unix timestamp at which the event was processed.
+	Time time.Time
+
+	// ExecutationTime is the amount of time needed for the command execution.
+	ExecutionTime time.Duration
+}
--- a/redis/redis.go
+++ b/redis/redis.go
@@ -190,8 +190,8 @@ type SlowLog struct {
 
 // Latency represents a redis LATENCY LATEST.
 type Latency struct {
-	// Latency event type.
-	EventType string
+	// Name of the latest latency spike event.
+	Name string
 
 	// Time of the latest latency spike for the event.
 	Time time.Time
--- a/redis/redis.go
+++ b/redis/redis.go
@@ -188,22 +188,22 @@ type SlowLog struct {
 	ClientName string
 }
 
-// Latency represents a redis LATENCY LATEST
+// Latency represents a redis LATENCY LATEST.
 type Latency struct {
-	// Latency event name
-	Event string
+	// Latency event type.
+	EventType string
 
-	// Time is the unix timestamp at which the event was processed.
+	// Time of the latest latency spike for the event.
 	Time time.Time
 
-	// LatestExecutationTime is the amount of time needed for the command execution.
-	LatestExecutionTime time.Duration
+	// Latest is the latest recorded latency for the named event.
+	Latest time.Duration
 
-	// Records the all-time maximum latency for every element
-	MaxExecutionTime time.Duration
+	// Max is the maximum latency for the named event.
+	Max time.Duration
 }
 
-// LatencyHistory represents a redis LATENCY HISTORY
+// LatencyHistory represents a redis LATENCY HISTORY.
 type LatencyHistory struct {
 	// Time is the unix timestamp at which the event was processed.
 	Time time.Time
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -646,8 +646,8 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
 	return logs, nil
 }
 
-// Latency is a helper that parse the LATENCY LATEST command output and
-// return the array of Latency
+// Latencies is a helper that parses the LATENCY LATEST command output and
+// return the slice of Latency values.
 func Latencies(result interface{}, err error) ([]Latency, error) {
 	rawLatencies, err := Values(result, err)
 	if err != nil {
@@ -665,7 +665,7 @@ func Latencies(result interface{}, err error) ([]Latency, error) {
 			return nil, fmt.Errorf("redigo: LATENCY LATEST element has %d elements, expected 4", len(rawLatency))
 		}
 
-		event.Event, err = String(rawLatency[0], nil)
+		event.EventType, err = String(rawLatency[0], nil)
 		if err != nil {
 			return nil, fmt.Errorf("redigo: LATENCY LATEST element[0] is not a string: %w", err)
 		}
@@ -682,14 +682,14 @@ func Latencies(result interface{}, err error) ([]Latency, error) {
 			return nil, fmt.Errorf("redigo: LATENCY LATEST element[2] not an int64, got %T", rawLatency[2])
 		}
 
-		event.LatestExecutionTime = time.Duration(latestDuration) * time.Millisecond
+		event.Latest = time.Duration(latestDuration) * time.Millisecond
 
 		maxDuration, ok := rawLatency[3].(int64)
 		if !ok {
 			return nil, fmt.Errorf("redigo: LATENCY LATEST element[3] not an int64, got %T", rawLatency[3])
 		}
 
-		event.MaxExecutionTime = time.Duration(maxDuration) * time.Millisecond
+		event.Max = time.Duration(maxDuration) * time.Millisecond
 
 		latencies[i] = event
 	}
@@ -697,7 +697,7 @@ func Latencies(result interface{}, err error) ([]Latency, error) {
 }
 
 // LatencyHistories is a helper that parse the LATENCY HISTORY command output and
-// return the array of LatencyHistory
+// return the array of LatencyHistory.
 func LatencyHistories(result interface{}, err error) ([]LatencyHistory, error) {
 	rawLogs, err := Values(result, err)
 	if err != nil {
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -663,7 +663,7 @@ func Latencies(result interface{}, err error) ([]Latency, error) {
 
 		var event Latency
 		if len(rawLatency) != 4 {
-			return nil, fmt.Errorf("redigo: LATENCY LATEST element has %d elements, expected 4", len(rawLatency))
+			return nil, fmt.Errorf("redigo: latencies element has %d elements, expected 4", len(rawLatency))
 		}
 
 		event.Name, err = String(rawLatency[0], nil)
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -723,7 +723,7 @@ func LatencyHistories(result interface{}, err error) ([]LatencyHistory, error) {
 
 		duration, ok := rawLog[1].(int64)
 		if !ok {
-			return nil, fmt.Errorf("redigo: LATENCY HISTORY element[1] not an int64, got %T", rawLog[1])
+			return nil, fmt.Errorf("redigo: latency history element[1] not an int64, got %T", rawLog[1])
 		}
 
 		event.ExecutionTime = time.Duration(duration) * time.Millisecond
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -668,7 +668,7 @@ func Latencies(result interface{}, err error) ([]Latency, error) {
 
 		event.Name, err = String(rawLatency[0], nil)
 		if err != nil {
-			return nil, fmt.Errorf("redigo: LATENCY LATEST element[0] is not a string: %w", err)
+			return nil, fmt.Errorf("redigo: latencies element[0] is not a string: %w", err)
 		}
 
 		timestamp, ok := rawLatency[1].(int64)
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -699,7 +699,7 @@ func Latencies(result interface{}, err error) ([]Latency, error) {
 }
 
 // LatencyHistories is a helper that parse the LATENCY HISTORY command output and
-// return the array of LatencyHistory.
+// returns a LatencyHistory slice.
 func LatencyHistories(result interface{}, err error) ([]LatencyHistory, error) {
 	rawLogs, err := Values(result, err)
 	if err != nil {
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -645,3 +645,88 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
 	}
 	return logs, nil
 }
+
+// Latency is a helper that parse the LATENCY LATEST command output and
+// return the array of Latency
+func Latencies(result interface{}, err error) ([]Latency, error) {
+	rawLatencies, err := Values(result, err)
+	if err != nil {
+		return nil, err
+	}
+	latencies := make([]Latency, len(rawLatencies))
+	for i, e := range rawLatencies {
+		rawLatency, ok := e.([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("redigo: rawLatency element is not an array, got %T", e)
+		}
+
+		var event Latency
+		if len(rawLatency) != 4 {
+			return nil, fmt.Errorf("redigo: LATENCY LATEST element has %d elements, expected 4", len(rawLatency))
+		}
+
+		event.Event, err = String(rawLatency[0], nil)
+		if err != nil {
+			return nil, fmt.Errorf("redigo: LATENCY LATEST element[0] is not a string: %w", err)
+		}
+
+		timestamp, ok := rawLatency[1].(int64)
+		if !ok {
+			return nil, fmt.Errorf("redigo: LATENCY LATEST element[1] not an int64, got %T", rawLatency[1])
+		}
+
+		event.Time = time.Unix(timestamp, 0)
+
+		latestDuration, ok := rawLatency[2].(int64)
+		if !ok {
+			return nil, fmt.Errorf("redigo: LATENCY LATEST element[2] not an int64, got %T", rawLatency[2])
+		}
+
+		event.LatestExecutionTime = time.Duration(latestDuration) * time.Millisecond
+
+		maxDuration, ok := rawLatency[3].(int64)
+		if !ok {
+			return nil, fmt.Errorf("redigo: LATENCY LATEST element[3] not an int64, got %T", rawLatency[3])
+		}
+
+		event.MaxExecutionTime = time.Duration(maxDuration) * time.Millisecond
+
+		latencies[i] = event
+	}
+	return latencies, nil
+}
+
+// LatencyHistories is a helper that parse the LATENCY HISTORY command output and
+// return the array of LatencyHistory
+func LatencyHistories(result interface{}, err error) ([]LatencyHistory, error) {
+	rawLogs, err := Values(result, err)
+	if err != nil {
+		return nil, err
+	}
+	latencyHistories := make([]LatencyHistory, len(rawLogs))
+	for i, e := range rawLogs {
+		rawLog, ok := e.([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("LATENCY HISTORY: latencyHistory element is not an array, got %T", e)
+		}
+
+		var event LatencyHistory
+
+		timestamp, ok := rawLog[0].(int64)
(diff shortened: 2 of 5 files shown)


redis/reply_test.go
```diff
@@ -225,6 +225,130 @@ func TestSlowLog(t *testing.T) {
 	}
 }
 
+func TestLatency(t *testing.T) {
+	c, err := dial()
+	if err != nil {
+		t.Errorf("TestLatency failed during dial with error " + err.Error())
+		return
+	}
```
stevenh, line 233: I would use `testify.require` which makes for a much simpler test flow and is already used in other places:
```suggestion
	c, err := dial()
	require.NoError(t, err)
```

Alternatively `T.Fatalf` achieves what you doing here in one line, with `f` methods it's idiomatic to use formatting parameters. The context will be apparent from the test failing, so no need to name the test for example:
```suggestion
	c, err := dial()
	if err != nil {
		t.Fatalf("dial failed: %s", err)
	}
```

redis/reply_test.go
```diff
@@ -225,6 +225,130 @@ func TestSlowLog(t *testing.T) {
...
+	defer c.Close()
+
+	resultStr, err := redis.Strings(c.Do("CONFIG", "GET", "latency-monitor-threshold"))
+	if err != nil {
+		t.Errorf("TestLatency failed during CONFIG GET latency-monitor-threshold with error " + err.Error())
+		return
+	}
```
stevenh, line 240: As above, more below.
```suggestion
	require.NoError(t, err)
```
alternative:
```suggestion
	if err != nil {
		t.Fatalf("CONFIG GET latency-monitor-threshold: %s", err)
	}
```

redis/reply_test.go
```diff
@@ -225,6 +225,130 @@ func TestSlowLog(t *testing.T) {
...
+	resultStr, err := redis.Strings(c.Do("CONFIG", "GET", "latency-monitor-threshold"))
+	if err != nil {
+		t.Errorf("TestLatencyHistories failed during CONFIG GET latency-monitor-threshold with error " + err.Error())
+		return
+	}
+	// in case of older verion < 2.8.13 where LATENCY command is not supported
+	// don't run the test
+	if len(resultStr) == 0 {
+		return
+	}
```
stevenh, line 308: There's two copies of this, you could make a little helper.

redis/reply_test.go
```diff
@@ -225,6 +225,130 @@ func TestSlowLog(t *testing.T) {
...
+		t.Errorf("%s=%+v, want %+v", "TestLatencyHistories test failed : ",
+			fmt.Sprint(len(latencyHistory))+" "+fmt.Sprint(latencyEvent.ExecutionTime), "1 1ms")
+	}
+
+	// reset the old configuration after test
+	result, err = c.Do("CONFIG", "SET", "latency-monitor-threshold", latencyMonitorThresholdOldCfg)
+	if err != nil && result != "OK" {
+		t.Errorf("TestLatencyHistories failed during CONFIG SET with error " + err.Error())
+		return
+	}
```
stevenh, line 349: Another little helper.

redis/redis.go
```diff
@@ -187,3 +187,27 @@ type SlowLog struct {
 	// ClientName is the name set via the CLIENT SETNAME command (4.0 only).
 	ClientName string
 }
+
+// Latency represents a redis LATENCY LATEST
+type Latency struct {
+	// Latency event name
+	Event string
+
+	// Time is the unix timestamp at which the event was processed.
```
stevenh, line 196: This isn't a unix timestamp, same below.
```suggestion
	// Time of the latest latency spike for the event.
```

redis/redis.go
```diff
@@ -187,3 +187,27 @@ type SlowLog struct {
 	// ClientName is the name set via the CLIENT SETNAME command (4.0 only).
 	ClientName string
 }
+
+// Latency represents a redis LATENCY LATEST
+type Latency struct {
+	// Latency event name
+	Event string
```
stevenh, line 194: Would this make more sense to call it Name?
```suggestion
	// Name is the name of event.
	Name string
```

redis/redis.go
```diff
@@ -187,3 +187,27 @@ type SlowLog struct {
 	// ClientName is the name set via the CLIENT SETNAME command (4.0 only).
 	ClientName string
 }
+
+// Latency represents a redis LATENCY LATEST
```
stevenh, line 191: Always end comments with a full stop.
```suggestion
// Latency represents a redis LATENCY LATEST.
```

redis/redis.go
```diff
@@ -187,3 +187,27 @@ type SlowLog struct {
...
+	Event string
+
+	// Time is the unix timestamp at which the event was processed.
+	Time time.Time
+
+	// LatestExecutationTime is the amount of time needed for the command execution.
+	LatestExecutionTime time.Duration
+
+	// Records the all-time maximum latency for every element
+	MaxExecutionTime time.Duration
```
stevenh, line 203: Similar to above keeping it simple for the user.
```suggestion
	// Latest is the latest recorded latency for the named event.
	Latest time.Duration

	// Max is the maximum latency for the named event.
	Max time.Duration
```

stevenh (changes requested):
Thanks for the PR, there's a few things I've noted below.

While I think this could be helpful I'm wondering how much use it would get and hence should we put these type of helpers in their own sub package to avoid polluting the main API?

Something like:
`redis/latency` which would provide calls like:
```go
latency.Latest(result interface{}, err error) ([]LatestEvent, error)
latency.History(result interface{}, err error) ([]HistoryEvent, error)
```
Thoughts?

dmitri-lerko:
Thank you @stevenh, a lot of valuable suggestions and learning here! I'll aim to go through it over next couple of days. I'll aim to apply the same changes to SlowLogs helper/tests which I've based my change on.

stevenh:
> Thank you @stevenh, a lot of valuable suggestions and learning here! I'll aim to go through it over next couple of days. I'll aim to apply the same changes to SlowLogs helper/tests which I've based my change on.

I would encourage you do the changes to SlowLogs in a different PR, to avoid confusing issues.

dmitri-lerko:
> Thanks for the PR, there's a few things I've noted below.
> 
> While I think this could be helpful I'm wondering how much use it would get and hence should we put these type of helpers in their own sub package to avoid polluting the main API?
> 
> Something like: `redis/latency` which would provide calls like:
> 
> ```go
> latency.Latest(result interface{}, err error) ([]LatestEvent, error)
> latency.History(result interface{}, err error) ([]HistoryEvent, error)
> ```
> 
> Thoughts?

When developing a tool to logout `slowlog` entries I have failed to discover that there was a `SlowLogs` helper already implemented and spent some time implementing my own. I concerned that if this change introduce new pattern of `latency.` then it will further hinder discoverability of the APIs.

```go
latency.Latest(result interface{}, err error) ([]LatestEvent, error)
latency.History(result interface{}, err error) ([]HistoryEvent, error)
```

redis/redis.go
dmitri-lerko: May-be even `EventType`?

dmitri-lerko:
I've applied suggested changes in all the instances that I could find. Tests look a lot more dense and readable. 

Not sure what is more usable:
```
	resultStr, err := redis.Strings(c.Do("CONFIG", "GET", "latency-monitor-threshold"))
	require.NoError(t, err, "TestLatency failed during CONFIG GET latency-monitor-threshold.")
```

or
```
	resultStr, err := redis.Strings(c.Do("CONFIG", "GET", "latency-monitor-threshold"))
	require.NoError(t, err, "TestLatency failed during CONFIG GET latency-monitor-threshold.")
```

Framework does point at the problematic test, but lacking of particular details outside of the error message:
```
=== RUN   TestLatency
    reply_test.go:231:
        	Error Trace:	reply_test.go:231
        	Error:      	Received unexpected error:
        	            	server exited: 96422:M 11 Jun 2022 11:06:36.347 # Failed listening on port 16111 (TCP), aborting.
        	Test:       	TestLatency
        	Messages:   	TestLatency failed during dial.
--- FAIL: TestLatency (0.04s)
```

Additionally, I've renamed `Event` to `EventType`, but also happy to just change it to `Name` as per your original suggestion.

stevenh:
> I've applied suggested changes in all the instances that I could find. Tests look a lot more dense and readable.
> 
> Not sure what is more usable:
> 
> ```
> 	resultStr, err := redis.Strings(c.Do("CONFIG", "GET", "latency-monitor-threshold"))
> 	require.NoError(t, err, "TestLatency failed during CONFIG GET latency-monitor-threshold.")
> ```
> 
> or
> 
> ```
> 	resultStr, err := redis.Strings(c.Do("CONFIG", "GET", "latency-monitor-threshold"))
> 	require.NoError(t, err)
> ```
> 
> Framework does point at the problematic test, but lacking of particular details outside of the error message:
> 
> ```
> === RUN   TestLatency
>     reply_test.go:231:
>         	Error Trace:	reply_test.go:231
>         	Error:      	Received unexpected error:
>         	            	server exited: 96422:M 11 Jun 2022 11:06:36.347 # Failed listening on port 16111 (TCP), aborting.
>         	Test:       	TestLatency
>         	Messages:   	TestLatency failed during dial.
> --- FAIL: TestLatency (0.04s)
> ```
> 
> Additionally, I've renamed `Event` to `EventType`, but also happy to just change it to `Name` as per your original suggestion.

My take on is that in 99% of cases you get all the context needed from the test name `TestLatency` and the `file:line` at which it failed `reply_test.go:231` so there's no need to add addition context.

If you added the message `TestLatency failed during CONFIG GET latency-monitor-threshold` and the function was renamed then the message is wrong, it's also redundant information as the test output already includes that, so I would never name the test in the message.

Adding `CONFIG GET latency-monitor-threshold` does inform the user what failed when the tests runs, however that is unlikely to help them fix the issue, that needs additional context of the surrounding code, which can already be found from `file:line`.

The way I look at it is use a message where it helps the developer understand why this check exists if it's not clear, which for almost all cases is already obvious particularly so for `error` check.

Another thing in that snippet is the variable naming `resultStr`. As a typed language it's not considered idiomatic to put type in the variable name unless you need to disambiguate with other variables. In this case it's also misleading as the result of `redis.Strings()` is actually a `slice` not a `string`. You should choose a name which adds enough context for the scope that is used. For example in this case I would just use `res` as it's only needed for one line below.

As a full example I would go for the following in this particualar case:
```go
	res, err := redis.Strings(c.Do("CONFIG", "GET", "latency-monitor-threshold"))
	require.NoError(t, err)
	if len(res) == 0 {
		t.Skip("Latency commands not supported, added in redis 2.8.13")
	}
```

Hope that helps.

redis/reply.go
```diff
@@ -645,3 +645,88 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
...
+}
+
+// LatencyHistories is a helper that parse the LATENCY HISTORY command output and
+// return the array of LatencyHistory.
+func LatencyHistories(result interface{}, err error) ([]LatencyHistory, error) {
+	rawLogs, err := Values(result, err)
+	if err != nil {
+		return nil, err
+	}
+	latencyHistories := make([]LatencyHistory, len(rawLogs))
```
stevenh, line 706:
```suggestion
	}

	latencyHistories := make([]LatencyHistory, len(rawLogs))
```

redis/reply.go
```diff
@@ -645,3 +645,88 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
...
+	for i, e := range rawLogs {
+		rawLog, ok := e.([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("LATENCY HISTORY: latencyHistory element is not an array, got %T", e)
+		}
+
+		var event LatencyHistory
+
+		timestamp, ok := rawLog[0].(int64)
```
stevenh, line 715:
```suggestion
		var event LatencyHistory
		timestamp, ok := rawLog[0].(int64)
```

redis/reply.go
```diff
@@ -645,3 +645,88 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
...
+		duration, ok := rawLog[1].(int64)
+		if !ok {
+			return nil, fmt.Errorf("redigo: LATENCY HISTORY element[1] not an int64, got %T", rawLog[1])
+		}
+
+		event.ExecutionTime = time.Duration(duration) * time.Millisecond
+
+		latencyHistories[i] = event
+	}
+	return latencyHistories, nil
```
stevenh, line 731:
```suggestion
	}

	return latencyHistories, nil
```

redis/reply_test.go
```diff
@@ -225,6 +226,81 @@ func TestSlowLog(t *testing.T) {
 	}
 }
 
+func TestLatency(t *testing.T) {
+	c, err := dial()
+	require.NoError(t, err, "TestLatency failed during dial.")
+	defer c.Close()
+
+	resultStr, err := redis.Strings(c.Do("CONFIG", "GET", "latency-monitor-threshold"))
+	require.NoError(t, err, "TestLatency failed during CONFIG GET latency-monitor-threshold.")
```
stevenh, line 235: Messages should not include the test name, more below and see comment thread for more details.

redis/test_test.go
```diff
@@ -39,7 +39,7 @@ var (
 
 	serverPath     = flag.String("redis-server", "redis-server", "Path to redis server binary")
 	serverAddress  = flag.String("redis-address", "127.0.0.1", "The address of the server")
-	serverBasePort = flag.Int("redis-port", 16379, "Beginning of port range for test servers")
+	serverBasePort = flag.Int("redis-port", 16112, "Beginning of port range for test servers")
```
stevenh, line 42: Please don't change this, it can be adjusted to your environment with the command line flag.

redis/reply_test.go
```diff
@@ -225,6 +226,81 @@ func TestSlowLog(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, "OK", result, "TestLatency failed during CONFIG SET")
+
+	// Sleep for 1ms to register a slow event.
+	_, err = c.Do("DEBUG", "SLEEP", 0.001)
+	require.NoError(t, err, "TestLatency failed during DEBUG SLEEP 0.001")
+	result, err = c.Do("LATENCY", "LATEST")
+	require.NoError(t, err, "TestLatency failed during LATENCY LATEST")
+	latestLatencies, err := redis.Latencies(result, err)
+	require.NoError(t, err, "TestLatency failed during redis.Latencies")
```
stevenh, line 253: Adding new lines between the steps can help improve the readability of tests as does excluding messages where they don't add value for the developer, for example:
```suggestion
	_, err = c.Do("DEBUG", "SLEEP", 0.001)
	require.NoError(t, err)

	result, err = c.Do("LATENCY", "LATEST")
	require.NoError(t, err)

	latestLatencies, err := redis.Latencies(result, err)
	require.NoError(t, err)
```

redis/reply_test.go
```diff
@@ -225,6 +226,81 @@ func TestSlowLog(t *testing.T) {
...
+
+	require.Equal(t, 1, len(latestLatencies))
+	latencyEvent := latestLatencies[0]
+	require.Equal(t, "command", latencyEvent.EventType)
+	require.Equal(t, time.Millisecond, latencyEvent.Latest)
+	require.Equal(t, time.Millisecond, latencyEvent.Max)
```
stevenh, line 259: You can combine this into one require which also makes it clearer too:
```suggestion
	expected := []Event{{
		EventType: "command",
		Latest: time.Millisecond,
		Max: time.Millisecond,
	}}
	require.Equal(t, latestLatencies ,expected)
```

redis/reply_test.go
```diff
@@ -225,6 +226,81 @@ func TestSlowLog(t *testing.T) {
...
+
+	// reset the old configuration after test.
+	result, err = c.Do("CONFIG", "SET", "latency-monitor-threshold", latencyMonitorThresholdOldCfg)
+	require.NoError(t, err)
+	require.Equal(t, "OK", result, "TestLatency failed during CONFIG SET")
```
stevenh, line 264: As any failure above would prevent this from being run this should be in a `defer` block directly after initial `SET` is validated.

redis/redis.go
```diff
@@ -187,3 +187,27 @@ type SlowLog struct {
 	// ClientName is the name set via the CLIENT SETNAME command (4.0 only).
 	ClientName string
 }
+
+// Latency represents a redis LATENCY LATEST.
+type Latency struct {
+	// Latency event type.
+	EventType string
```
stevenh, line 194: I liked this change, but then I tried to write the comment starting with `EventType` I couldn't make a good sentence without it stuttering, do you have anything or would this work?
```suggestion
	// Name of the latest latency spike event.
	Name string
```

Then I read the [redis LATENCY LATEST](https://redis.io/commands/latency-latest/) docs and it refers to the field as name
> Event name.
Unix timestamp of the latest latency spike for the event.
Latest event latency in millisecond.
All-time maximum latency for this event.

dmitri-lerko: Good points all around with REDIS docs being final arbiter here. Fully agree with a name being a better option.

redis/test_test.go
dmitri-lerko: Good spot, this was an unintended change for local testing that got accidentally added in.

redis/reply_test.go
dmitri-lerko: It would be a great to make this part more readable, however, in the suggestion we are also validating equality of Time which we can't predict in advance.

I've taken a hybrid approach where I am setting expected.Time from the latencyEvent.Time. This way we can collapse 3 requires into 1.

redis/reply_test.go
dmitri-lerko: Do we simply ignore return in this case? `defer c.Do("CONFIG", "SET", "latency-monitor-threshold", latencyMonitorThresholdOldCfg)`

dmitri-lerko:
@stevenh , thank you for the time invested into this. I've learned a lot already and made another iteration incorporating your feedback. Let  me know if there is anything else I can improve upon.

Thanks

stevenh:
Thanks @dmitri-lerko if you could mark all the comments you have addressed at resolved, that will make it easier to do another review :)

dmitri-lerko:
Hi @stevenh, I've resolved all but one comments. Please let me know what you think.

Thanks
Dmitri

redis/reply.go
```diff
@@ -645,3 +645,91 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
...
+		event.Max = time.Duration(maxDuration) * time.Millisecond
+
+		latencies[i] = event
+	}
+
+	return latencies, nil
+}
+
+// LatencyHistories is a helper that parse the LATENCY HISTORY command output and
+// return the array of LatencyHistory.
```
stevenh, line 702:
```suggestion
// returns a LatencyHistory slice.
```

redis/reply.go
```diff
@@ -645,3 +645,91 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
...
+	rawLatencies, err := Values(result, err)
+	if err != nil {
+		return nil, err
+	}
+
+	latencies := make([]Latency, len(rawLatencies))
+	for i, e := range rawLatencies {
+		rawLatency, ok := e.([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("redigo: rawLatency element is not an array, got %T", e)
```
stevenh, line 661: Errors should be lower case
```suggestion
			return nil, fmt.Errorf("redigo: latencies element is not slice, got %T", e)
```

redis/reply.go
```diff
@@ -645,3 +645,91 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
...
+		}
+
+		var event Latency
+		if len(rawLatency) != 4 {
+			return nil, fmt.Errorf("redigo: LATENCY LATEST element has %d elements, expected 4", len(rawLatency))
```
stevenh, line 666:
```suggestion
			return nil, fmt.Errorf("redigo: latencies element has %d elements, expected 4", len(rawLatency))
```

redis/reply.go
```diff
@@ -645,3 +645,91 @@ func SlowLogs(result interface{}, err error) ([]SlowLog, error) {
...
+		timestamp, ok := rawLatency[1].(int64)
+		if !ok {
+			return nil, fmt.Errorf("redigo: LATENCY LATEST element[1] not an int64, got %T", rawLatency[1])
+		}
+
+		event.Time = time.Unix(timestamp, 0)
+
+		latestDuration, ok := rawLatency[2].(int64)
+		if !ok {
+			return nil, fmt.Errorf("redigo: LATENCY LATEST element[2] not an int64, got %T", rawLatency[2])
```
stevenh, line 683:
```suggestion
			return nil, fmt.Errorf("redigo: latencies element[2] not an int64, got %T", rawLatency[2])
```

stevenh (changes requested):
Just some error message and comments to clean up and should be good to go.

stevenh:
Thanks for PR @dmitri-lerko much appreciated.

dmitri-lerko:
Hi @stevenh , thank you for your time on this PR. I'll aim to bring slowlog in light with the changes above to have better consistency.

PR #618, opened 2022-06-13
add support BytesMap reply.

zkep:


redis/reply_test.go
```diff
@@ -333,3 +333,40 @@ func ExampleString() {
 	// Output:
 	// "world"
 }
+
+func ExampleBytesMap() {
```
stevenh, line 337: Could you clarify why would you not just use `Values(...)` or `ScanStruct` a for this?

redis/reply.go
```diff
@@ -504,6 +504,26 @@ func Float64Map(result interface{}, err error) (map[string]float64, error) {
 	return m, err
 }
 
+// BytesMap is a helper that converts an array of strings (alternating key, value)
```
stevenh, line 507: Not sure what the common use case for this is, the example seems to indicated it would better to use `Values(..)` instead?

stevenh, line 507: Not sure what the common use case for this is, the example seems to indicated it would better to use `Values(..)` instead?

redis/reply_test.go
```diff
@@ -89,9 +89,9 @@ var replyTests = []struct {
 		ve([]string{"v1", "v2"}, nil),
 	},
 	{
-		"strings([string, string])",
-		ve(redis.Strings([]interface{}{"v1", "v2"}, nil)),
-		ve([]string{"v1", "v2"}, nil),
+		"stringMap([string, string])",
+		ve(redis.StringMap([]interface{}{[]byte("key1"), []byte("v1"), []byte("key2"), []byte("v2")}, nil)),
+		ve(map[string]string{"key1": "v1", "key2": "v2"}, nil),
```
stevenh, line 94: This looks like a mistake, as it's replacing a `Strings` test with `StringMap` test and not adding `BytesMap`

stevenh, line 94: This looks like a mistake, as it's replacing a `Strings` test with `StringMap` test and not adding `BytesMap`

stevenh (changes requested):
Not sure I understand the need here as it looks like `Values(...)` already does what you need?

PR #619, opened 2022-06-17
add support valueMap reply

zkep:
get map  form redis reply  this is often used

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -504,6 +504,22 @@ func Float64Map(result interface{}, err error) (map[string]float64, error) {
 	return m, err
 }
 
+// ValueMap is a helper that converts an array of strings (alternating key, value)
+// into a map[string]interface{}. The HGETALL commands return replies in this format.
+// Requires an even interface{} of values in result.
+func ValueMap(result interface{}, err error) (map[string]interface{}, error) {
+	var m map[string]interface{}
+	err = mapHelper(result, err, "ValueMap",
+		func(n int) {
+			m = make(map[string]interface{}, n)
+		}, func(key string, v interface{}) error {
+			m[key] = v
+			return nil
+		},
+	)
+	return m, err
+}
+
 // Positions is a helper that converts an array of positions (lat, long)
 // into a [][2]float64. The GEOPOS command returns replies in this format.
 func Positions(result interface{}, err error) ([]*[2]float64, error) {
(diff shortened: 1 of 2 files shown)


redis/reply.go
```diff
@@ -504,6 +504,22 @@ func Float64Map(result interface{}, err error) (map[string]float64, error) {
 	return m, err
 }
 
+// ValueMap is a helper that converts an array of strings (alternating key, value)
+// into a map[string]interface{}. The HGETALL commands return replies in this format.
+// Requires an even interface{} of values in result.
```
stevenh, line 509: I believe you mean:
```suggestion
// Requires an even number of values in result.
```

redis/reply_test.go
```diff
@@ -138,6 +138,11 @@ var replyTests = []struct {
 		ve(redis.Uint64(int64(-1), nil)),
 		ve(uint64(0), redis.ErrNegativeInt(-1)),
 	},
+	{
+		"valueMap([[]byte, []byte])",
+		ve(redis.ValueMap([]interface{}{[]byte("key1"), []byte("v1"), []byte("key2"), []byte("v2")}, nil)),
+		ve(map[string]interface{}{"key1": []byte("v1"), "key2": []byte("v2")}, nil),
```
stevenh, line 144: This isn't excessing `interface{}` as all the values are of the type `[]byte` which I don't believe is what the caller will expect.

stevenh (changes requested):

redis/reply_test.go
zkep: I want to get valueMap . then to individually process each  different types  key , such as  name,age, adult  etc...

`	_, err = c.Do("HMSET", "redis", "name", "go", "age", 18, "adult", true)
	if err != nil {
		fmt.Println(err)
		return
	}
	values, err := redis.ValueMap(c.Do("HGETALL", "redis"))
	if err != nil {
		fmt.Println(err)
		return
	}
	if _, ok := values["name"]; ok {
		value, _ := redis.String(values["name"], err)
		fmt.Printf("%#v\n", value)
	}
	if _, ok := values["age"]; ok {
		value, _ := redis.Int(values["age"], err)
		fmt.Printf("%#v\n", value)
	}
	if _, ok := values["adult"]; ok {
		value, _ := redis.Bool(values["adult"], err)
		fmt.Printf("%#v\n", value)
	}`

PR #620, opened 2022-06-28
fix: respect ctx in cmds run within DialContext

mitchsw:
The `DialContext()` function does not use the new `DoContext()` methods (added in v1.8.6) when configuring auth/clientName/db during connection creation. If the Redis server is unresponsive, this can cause `DialContext()` to run for a very long time (instead of cancelling when the deadline is exceeded). `DoContext` should be used for these cmds.

Repro steps:

1. Make Redis unresponsive: `CLIENT PAUSE 36000000 ALL`
2. Attempt to `DialContext()` with a 1sec timeout
3. Command is unexpectedly blocked for a long time

### DIFF
--- a/redis/conn.go
+++ b/redis/conn.go
@@ -291,21 +291,21 @@ func DialContext(ctx context.Context, network, address string, options ...DialOp
 			authArgs = append(authArgs, do.username)
 		}
 		authArgs = append(authArgs, do.password)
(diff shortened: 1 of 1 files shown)


mitchsw:
Thanks @stevenh! Is it possible to please merge & tag this fix?

stevenh:
To make sure you didn't miss it tag was done last week @mitchsw

PR #624, opened 2022-07-05
Prevent recursing infinitely and more tests

lwaddicor:
This prevents recursion without panicking and adds a few more tests


stevenh:
Closing as discussed.

Issue #626, opened 2022-07-14
Please don't let context pollute the apis


elgs:
I switched from go-redis because their apis are heavily polluted by the context. I come here to beg you please please don't pollute your apis with the context virus.

elgs:
@stevenh thank you for allowing your clients to not get involved in the context and still to get things done. 

I am not objecting context per se. The idea of context itself is nothing wrong. Since the http server is the first class citizen in Go, I understand context was built in the language by some opinionated people. That's ok. Context is no more than notifying some channels when things get canceled or timed out. In my opinion these are implementation details and are easy to be done in other/simpler ways. So I don't see why they need to be exposed to the api if it's not absolutely necessary. Also I don't believe context is the only most elegant way to achieve handling canceling and timeout. I have no problem anyone uses it for their implementation, but it's unnecessary to force the client to get involved, especially when they don't necessarily need it.

stevenh:
Thanks for the context @elgs.

I would say that the consistency that context brings with handling timeout and cancellation is quite a benefit.

It's all to common for API's to ignore the edge cases which are often inherited from the underlying transports such as TCP, which can result in some serious challenges when those rare cases occur, for example if keep-alive isn't enabled broken connections go unnoticed creating hard to detect resource leaks.

Context not only provides a consistent way of handling that, it does so in a way that makes the developer think about some of these edge cases.

Hopefully that provides some food for thought.

PR #629, opened 2022-07-25
 feat: add GetWithContext in pool, returns ConnWithContext

zhwei820:
~~While `activeConn` implements `ConnWithContext`, I think it's better to return as ConnWithContext.~~

~~Besides, `ConnWithContext` is a superset of `Conn`, `pool.Get` remains unchanged and works perfectly.~~

---

Hi, in consideration of **API compatibility**, I added a new method `GetWithContext` in pool instance, returns `ConnWithContext`, remaining others untouched. 

Fixed a bug in older tests code, and completed test code with `GetWithContext` method.

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -264,6 +264,22 @@ func (p *Pool) GetContext(ctx context.Context) (Conn, error) {
 	return &activeConn{p: p, pc: &poolConn{c: c, created: nowFunc()}}, nil
 }
 
+// GetWithContext gets a connection-with-context using the provided context.
+//
+// The provided Context must be non-nil. If the context expires before the
+// connection is complete, an error is returned. Any expiration on the context
+// will not affect the returned connection.
+//
+// If the function completes without error, then the application must close the
+// returned connection.
+func (p *Pool) GetWithContext(ctx context.Context) (ConnWithContext, error) {
+	if conn, err := p.GetContext(ctx); err != nil {
+		return nil, err
+	} else {
+		return conn.(ConnWithContext), nil // struct behind is activeConn, it implements ConnWithContext.
+	}
+}
+
 // PoolStats contains pool statistics.
 type PoolStats struct {
 	// ActiveCount is the number of connections in the pool. The count includes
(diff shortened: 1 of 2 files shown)


stevenh (commented):
While it's true that the returned type does implement `ConnWithContext` changing the return would break API compatibility for consumers.

For example if I had a interface which represented a pool including this method, that interface would not match requiring the consumer to update their code.

As such this type of change could only be introduced in a v2 release I'm afraid.

antsbean:
> 
In order to solve this problem ASAP, I suggest adding a new API GetContextV2 that doesn't  break API compatibility

zhwei820:
> While it's true that the returned type does implement `ConnWithContext` changing the return would break API compatibility for consumers.
> 
> For example if I had a interface which represented a pool including this method, that interface would not match requiring the consumer to update their code.
> 
> As such this type of change could only be introduced in a v2 release I'm afraid.

Hi, @stevenh, I renewed this PR. You may check it later.

stevenh (commented):
While better context support is something that could be considered as mentioned in a v2 API, adding now seems unnecessary is the consumer can solve this easily with a little wrapper.

During that consideration there has been a recent [request to not pollute the API with context](https://github.com/gomodule/redigo/issues/626).

Here's a simple wrapper that would enable what I believe you're trying to achieve with no library changes.
```go
type ctxPool struct {
    *redis.Pool
}

func (p *ctxPool) GetContext(ctx context.Context) (ConnWithContext, error) {
    c, err := p.Pool.GetContext(ctx)
    if err != nil {
        return c, err
    }

    if cc, ok := c.(ConnWithContext); ok {
        return cc, nil
    }

    return c, errors.New("connection does not support ConnWithTimeout")
}
```

That said you can call context methods without any of this as they validiate that the passed `Conn` supports `ConnWithContext` for example:
```go
c, err := p.GetContext(ctx)
if err != nil {
    return err
}

v, err := redis.DoContext(c, "GET", "mykey")
if err != nil {
   return err
}

fmt.Println(v)
```

zhwei820:
**adding now seems unnecessary is the consumer can solve this easily with a little wrapper.** Closed.

zhwei820:
Closed

zhwei820:
In the end, thank you for your maintenance of this awesome repo, and for quick and patient response. @stevenh   👍 👍

stevenh:
No problem @zhwei820 thanks for the PR, hope you now have a good way forward.

PR #649, opened 2023-06-07
fix: catch array value in stringMap

sergelogvinov:
Hello,

I've got the error `redigo: StringMap for "tls-allowlist" not a bulk string value, got []interface {}` using the StringMap func.

KeyDB has value-array in `CONFIG GET *` request.

What do you think, is ti possible to skip such value?
Or convert (join) array as string with `, `?

Thanks.

### DIFF
--- a/redis/reply.go
+++ b/redis/reply.go
@@ -421,12 +421,14 @@ func StringMap(reply interface{}, err error) (map[string]string, error) {
 		func(n int) {
 			result = make(map[string]string, n)
 		}, func(key string, v interface{}) error {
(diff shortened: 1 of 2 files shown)


stevenh (commented):
Thanks for the PR, unfortunately this changes the behaviour of StringMap which we can't / shouldn't do.

If KeyDB behaves differently there's nothing to stop you creating a custom function that correctly processes its results.

The comment in the docs, is there just to highlight one example of it's use.

sergelogvinov:
Yep, you are right. 
Thank you for feedback.

PR #654, opened 2023-10-10
Run the redis test server with enable-debug-command

elboulangero:
This is required for Redis >= 7.x, cf. https://raw.githubusercontent.com/redis/redis/7.0/00-RELEASENOTES:

```
* MODULE and DEBUG commands disabled (protected) by default, for better security (#9920)
```

Without this patch, some unit tests fail:

```
=== RUN   TestLatency
    reply_test.go:256:
                Error Trace:    /<<PKGBUILDDIR>>/_build/src/github.com/gomodule/redigo/redis/reply_test.go:256
                Error:          Received unexpected error:
                                ERR DEBUG command not allowed. If the enable-debug-command option is set to "local", you can run it from a local connection, otherwise you need to set this option in the configuration file, and then restart the server.
                Test:           TestLatency
--- FAIL: TestLatency (0.00s)
=== RUN   TestLatencyHistories
    reply_test.go:304:
                Error Trace:    /<<PKGBUILDDIR>>/_build/src/github.com/gomodule/redigo/redis/reply_test.go:304
                Error:          Received unexpected error:
                                ERR DEBUG command not allowed. If the enable-debug-command option is set to "local", you can run it from a local connection, otherwise you need to set this option in the configuration file, and then restart the server.
                Test:           TestLatencyHistories
--- FAIL: TestLatencyHistories (0.00s)
```

I tried to cook a clean patch but I'm not very familiar with Go, so it's not perfect.

### DIFF
--- a/redis/test_test.go
+++ b/redis/test_test.go
@@ -78,11 +78,11 @@ func redisServerVersion() (*Version, error) {
 		return nil, fmt.Errorf("invalid major version %q", match[1])
 	}
 
-	if v.minor, err = strconv.Atoi(match[1]); err != nil {
+	if v.minor, err = strconv.Atoi(match[2]); err != nil {
 		return nil, fmt.Errorf("invalid minor version %q", match[2])
 	}
 
-	if v.patch, err = strconv.Atoi(match[1]); err != nil {
+	if v.patch, err = strconv.Atoi(match[3]); err != nil {
 		return nil, fmt.Errorf("invalid patch version %q", match[3])
 	}
 
--- a/redis/test_test.go
+++ b/redis/test_test.go
@@ -60,43 +60,33 @@ type version struct {
 	patch int
 }
 
-func GetRedisVersion() (*Version, error) {
+func redisServerVersion() (*Version, error) {
 	out, err := exec.Command("redis-server", "--version").Output()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("server version: %w", err)
 	}
 
-	version := ""
-	fields := strings.Fields(string(out))
-	for _, field := range fields {
-		if strings.HasPrefix(field, "v=") {
-			version = field[2:]
-			break
-		}
-	}
-	if version == "" {
-		return nil, errors.New("no version string")
+	re := regexp.MustCompile(`v=(\d+)\.(\d+)\.(\d+)`)
+	match := re.FindStringSubmatch(ver)
+	if len(match) != 4 {
+		return nil, fmt.Errorf("no server version found in %q", ver)
 	}
 
-	values := [3]int{}
-	elems := strings.Split(version, ".")
-	for i, _ := range elems {
-		if i > 2 {
-			break
-		}
-		values[i], err = strconv.Atoi(elems[i])
-		if err != nil {
-			return nil, err
-		}
+	var err error
+	var v version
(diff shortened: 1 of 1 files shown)


redis/test_test.go
```diff
@@ -54,7 +54,61 @@ type Server struct {
 	done chan struct{}
 }
 
+type Version struct {
+	major int
+	minor int
+	micro int
+}
```
stevenh, line 61: As this not needed outside the package there's no need to export it so lower case the type name.

Also switching to semantic version naming of the parameters.
```suggestion
type version struct {
	major int
	minor int
	patch int
}
```

redis/test_test.go
```diff
@@ -54,7 +54,61 @@ type Server struct {
...
+	}
+
+	v := &Version{
+		major: values[0],
+		minor: values[1],
+		micro: values[2],
+	}
+
+	return v, nil
+}
```
stevenh, line 100: This can be simplified to the following. Notice a few things:
* It's idiomatic in go to avoid naming things `get` / `set`
* Wrap errors with `fmt.Errorf` to provide context to the user
```suggestion
func redisServerVersion() (*Version, error) {
	out, err := exec.Command("redis-server", "--version").Output()
	if err != nil {
		return nil, fmt.Errorf("server version: %w", err)
	}

	re := regexp.MustCompile(`v=(\d+)\.(\d+)\.(\d+)`)
	match := re.FindStringSubmatch(ver)
	if len(match) != 4 {
		return nil, fmt.Errorf("no server version found in %q", ver)
	}

	var err error
	var v version
	if v.major, err = strconv.Atoi(match[1]); err != nil {
		return nil, fmt.Errorf("invalid major version %q", match[1])
	}

	if v.minor, err = strconv.Atoi(match[1]); err != nil {
		return nil, fmt.Errorf("invalid minor version %q", match[2])
	}

	if v.patch, err = strconv.Atoi(match[1]); err != nil {
		return nil, fmt.Errorf("invalid patch version %q", match[3])
	}

	return &v, nil
}
```

redis/test_test.go
```diff
@@ -54,7 +54,61 @@ type Server struct {
...
+
 func NewServer(name string, args ...string) (*Server, error) {
+	version, err := GetRedisVersion()
```
stevenh, line 103:
```suggestion
	version, err := redisServerVersion()
```

stevenh (changes requested):
Thanks for PR, appreciated.

As you said your new to go, I've taken the opportunity to share some improvements to this, hope that helps.

redis/test_test.go
elboulangero: Little typo above, `strconv.Atoi(match[1])` repeated 3 times without incrementing the index.

elboulangero: I don't know how to modify, so I'll commit the suggestion and edit it afterward.

redis/test_test.go
elboulangero: test: GetRedisVersion -> redisServerVersion

elboulangero:
Thanks for the review and advice!

I applied your suggestions, then fixed the details to get it to build. I think it's good now.

stevenh:
Sorry @elboulangero missed the notification for this, merged.

elboulangero:
@stevenh no worries!

PR #660, opened 2024-02-11
feat: add TestOnBorrowContext

vasayxtx:
Add TestOnBorrowContext to the Pool struct for checking the health of the idle connection with a given context.

### DIFF
--- a/redis/pool.go
+++ b/redis/pool.go
@@ -128,16 +128,24 @@ type Pool struct {
 	// DialContext is an application supplied function for creating and configuring a
 	// connection with the given context.
 	//
-	// The connection returned from Dial must not be in a special state
+	// The connection returned from DialContext must not be in a special state
 	// (subscribed to pubsub channel, transaction started, ...).
 	DialContext func(ctx context.Context) (Conn, error)
 
 	// TestOnBorrow is an optional application supplied function for checking
 	// the health of an idle connection before the connection is used again by
-	// the application. Argument t is the time that the connection was returned
+	// the application. Argument lastUsed is the time when the connection was returned
 	// to the pool. If the function returns an error, then the connection is
 	// closed.
-	TestOnBorrow func(c Conn, t time.Time) error
+	TestOnBorrow func(c Conn, lastUsed time.Time) error
+
+	// TestOnBorrowContext is an optional application supplied function
+	// for checking the health of an idle connection with the given context
+	// before the connection is used again by the application.
+	// Argument lastUsed is the time when the connection was returned
+	// to the pool. If the function returns an error, then the connection is
+	// closed.
+	TestOnBorrowContext func(ctx context.Context, c Conn, lastUsed time.Time) error
 
 	// Maximum number of idle connections in the pool.
 	MaxIdle int
@@ -228,6 +236,7 @@ func (p *Pool) GetContext(ctx context.Context) (Conn, error) {
 		p.idle.popFront()
 		p.mu.Unlock()
 		if (p.TestOnBorrow == nil || p.TestOnBorrow(pc.c, pc.t) == nil) &&
+			(p.TestOnBorrowContext == nil || p.TestOnBorrowContext(ctx, pc.c, pc.t) == nil) &&
 			(p.MaxConnLifetime == 0 || nowFunc().Sub(pc.created) < p.MaxConnLifetime) {
 			return &activeConn{p: p, pc: pc}, nil
 		}
--- a/redis/pool_test.go
+++ b/redis/pool_test.go
@@ -48,6 +48,25 @@ func (c *poolTestConn) Close() error {
 func (c *poolTestConn) Err() error { return c.err }
 
 func (c *poolTestConn) Do(commandName string, args ...interface{}) (interface{}, error) {
+	return c.do(c.Conn.Do, commandName, args...)
+}
+
+func (c *poolTestConn) DoContext(ctx context.Context, commandName string, args ...interface{}) (interface{}, error) {
+	cwc, ok := c.Conn.(redis.ConnWithContext)
+	if !ok {
+		return nil, errors.New("redis: connection does not support ConnWithContext")
+	}
+	return c.do(
+		func(c string, a ...interface{}) (interface{}, error) {
+			return cwc.DoContext(ctx, c, a...)
+		},
+		commandName, args)
+}
+
+func (c *poolTestConn) do(
+	fn func(commandName string, args ...interface{}) (interface{}, error),
+	commandName string, args ...interface{},
+) (interface{}, error) {
 	if commandName == "ERR" {
 		c.err = args[0].(error)
 		commandName = "PING"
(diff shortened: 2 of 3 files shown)


redis/pool_test.go
```diff
@@ -48,21 +48,46 @@ func (c *poolTestConn) Close() error {
...
+}
+
+func (c *poolTestConn) DoContext(ctx context.Context, commandName string, args ...interface{}) (interface{}, error) {
+	cwt, ok := c.Conn.(redis.ConnWithContext)
+	if !ok {
+		return nil, errors.New("redis: connection does not support ConnWithContext")
+	}
+	return c.do(
+		func(c string, a ...interface{}) (interface{}, error) { return cwt.DoContext(ctx, c, a...) },
+		commandName, args)
```
stevenh, line 61: Could we expand this out to make it easier to read, assumes previous change too.
```suggestion
	return c.do(
		func(command string, args ...interface{}) (interface{}, error) {
			return cwc.DoContext(ctx, command, args...)
		},
		commandName, args
	)
```

redis/pool_test.go
```diff
@@ -48,21 +48,46 @@ func (c *poolTestConn) Close() error {
 func (c *poolTestConn) Err() error { return c.err }
 
 func (c *poolTestConn) Do(commandName string, args ...interface{}) (interface{}, error) {
+	return c.do(c.Conn.Do, commandName, args...)
+}
+
+func (c *poolTestConn) DoContext(ctx context.Context, commandName string, args ...interface{}) (interface{}, error) {
+	cwt, ok := c.Conn.(redis.ConnWithContext)
```
stevenh, line 55: I know this is what the rest of the code base has but it's actually a cut ant past error from ConnWithTimeout, can we change this to use `cwc` so it correctly represents the actual type.
```suggestion
	cwc, ok := c.Conn.(redis.ConnWithContext)
```

redis/pool_test.go
```diff
@@ -48,21 +48,46 @@ func (c *poolTestConn) Close() error {
...
 	c.d.commands = append(c.d.commands, commandName)
 	return c.Conn.Send(commandName, args...)
 }
 
+func (c *poolTestConn) ReceiveContext(ctx context.Context) (reply interface{}, err error) {
+	cwt, ok := c.Conn.(redis.ConnWithContext)
+	if !ok {
+		return nil, errors.New("redis: connection does not support ConnWithContext")
+	}
+	return cwt.ReceiveContext(ctx)
```
stevenh, line 88: As above `cwc`
```suggestion
	cwc, ok := c.Conn.(redis.ConnWithContext)
	if !ok {
		return nil, errors.New("redis: connection does not support ConnWithContext")
	}

	return cwc.ReceiveContext(ctx)
```

redis/pool_test.go
```diff
@@ -90,15 +119,14 @@ func (d *poolDialer) dial() (redis.Conn, error) {
 	return &poolTestConn{d: d, Conn: c}, nil
 }
 
-func (d *poolDialer) dialContext(ctx context.Context) (redis.Conn, error) {
-	return d.dial()
-}
-
 func (d *poolDialer) check(message string, p *redis.Pool, dialed, open, inuse int) {
+	d.t.Helper()
```
stevenh, line 123: Nice catch!

redis/pool_test.go
```diff
@@ -368,21 +396,92 @@ func TestPoolConcurrenSendReceive(t *testing.T) {
...
-	defer p.Close()
 
-	for i := 0; i < 10; i++ {
-		c := p.Get()
-		_, err := c.Do("PING")
-		require.NoError(t, err)
-		c.Close()
-	}
-	d.check("1", p, 10, 1, 0)
+	t.Run("TestOnBorrow returns error", func(t *testing.T) {
```
stevenh, line 409: Can we keep the test names short without spaces yet still descriptive e.g. `TestOnBorrow-error` this makes the names searchable and the output concise.

redis/test_test.go
```diff
@@ -197,15 +198,22 @@ func DefaultServerAddr() (string, error) {
 // DialDefaultServer starts the test server if not already started and dials a
 // connection to the server.
 func DialDefaultServer(options ...DialOption) (Conn, error) {
+	return DialDefaultServerContext(context.Background(), options...)
+}
+
+// DialDefaultServerContext starts the test server if not already started and
+// dials a connection to the server. The provided context is used for the dial
+// and for the initial FLUSHDB command.
```
stevenh, line 206: Lets avoid describing internal detail, so we don't have to update if this changes.
```suggestion
// DialDefaultServerContext starts the test server if not already started and
// dials a connection to the server.
```

redis/pool_test.go
```diff
@@ -368,21 +396,92 @@ func TestPoolConcurrenSendReceive(t *testing.T) {
...
-		MaxIdle:      2,
-		Dial:         d.dial,
-		TestOnBorrow: func(redis.Conn, time.Time) error { return redis.Error("BLAH") },
+	pingN := func(ctx context.Context, p *redis.Pool, n int) {
+		for i := 0; i < n; i++ {
+			c, err := p.GetContext(ctx)
+			require.NoError(t, err)
+			_, err = redis.DoContext(c, ctx, "PING")
+			require.NoError(t, err)
+			require.NoError(t, c.Close())
```
stevenh, line 405: defer the close so it will always be executed even if one of the other tests fail.
```suggestion
			c, err := p.GetContext(ctx)
			require.NoError(t, err)
			defer func() {
				require.NoError(t, c.Close())
			}()

			_, err = redis.DoContext(c, ctx, "PING")
			require.NoError(t, err)
```

stevenh (changes requested):
Thanks for the PR appreciated.

I've done a quick pass, some minor nits to clean up, will do another pass once those are addressed.

redis/pool_test.go
vasayxtx: sure, no problem.

redis/pool_test.go
vasayxtx: Ok, done.

vasayxtx:
@stevenh, thanks, all comments are processed. Please review again:)

stevenh:
Would be great to get your input on #661 when you have a min @vasayxtx

stevenh (changes requested):
So apparently I didn't hit submit on this, so apologies its a been a while.

redis/pool.go
```diff
@@ -139,6 +139,14 @@ type Pool struct {
 	// closed.
 	TestOnBorrow func(c Conn, t time.Time) error
 
+	// TestOnBorrowContext is an optional application supplied function
+	// for checking the health of an idle connection with the given context
+	// before the connection is used again by the application.
+	// Argument t is the time that the connection was returned
+	// to the pool. If the function returns an error, then the connection is
+	// closed.
+	TestOnBorrowContext func(ctx context.Context, c Conn, t time.Time) error
```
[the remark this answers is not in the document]
vasayxtx: Good point, renamed for both `TestOnBorrowContext` and  `TestOnBorrow`. Additionally, started to check the `lastUsed` arg in tests.

stevenh:
Thanks for your work on this @vasayxtx appreciated!

vasayxtx:
@stevenh, no problem, thank you for the review!:)

PR #686, opened 2025-03-12
fix: honor server-path in test_test.go

elboulangero:
So that it's possible to run tests with:

```
go test [...] -redis-server valkey-server
```


stevenh (approved):
Nice catch

